Issue #2, opened 2020-01-13
Filtering Classes by teacher in Admin displays all users


geordi:
When you want to filter Classes in Admin by a teacher, the filtering box contains all the users. Only teachers should be displayed.

PR #5, opened 2020-01-14
Fix: Filter by teacher in admin

geordi:
Displays only teacher logins in the filter box. Fixes Issue #2

### DIFF
--- a/common/admin.py
+++ b/common/admin.py
@@ -18,7 +18,7 @@ class BaseByTeacherFilter(admin.SimpleListFilter):
         teachers = User.objects.filter(groups__name='teachers')
         items = ( (t.id, t.username) for t in teachers )
 
-        return tuple(items)
+        return items
 
     def queryset(self, request, queryset):
         pass
--- a/common/admin.py
+++ b/common/admin.py
@@ -4,10 +4,40 @@ from django.contrib.auth.models import User, Group
 from django.contrib.auth.admin import UserAdmin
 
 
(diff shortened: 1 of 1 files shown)


common/admin.py
```diff
@@ -4,10 +4,40 @@
...
+    parameter_name = 'teacher'
+
+    def lookups(self, request, model_admin):
+        teachers = User.objects.filter(groups__name='teachers')
+        items = ( (t.id, t.username) for t in teachers )
+
+        return tuple(items)
+
+    def queryset(self, request, queryset):
+        pass
```
trnila: it can be omitted

common/admin.py
```diff
@@ -4,10 +4,40 @@
...
+    and https://stackoverflow.com/questions/880684/in-django-admin-how-to-filter-users-by-group
+    """
+    title = 'teacher'
+    parameter_name = 'teacher'
+
+    def lookups(self, request, model_admin):
+        teachers = User.objects.filter(groups__name='teachers')
+        items = ( (t.id, t.username) for t in teachers )
+
+        return tuple(items)
```
trnila: `items` is already a tuple, so it can be returned immediately

geordi: Well, `items` is a generator and I didn't know it'd work. I can definitely test it.

common/admin.py
geordi: Wanted to be explicit.

PR #115, opened 2021-05-06
Use BytesIO instead of StringIO in test files

Kobzol:


evaluator/testsets.py
```diff
@@ -16,14 +16,14 @@ def __init__(self, path):
         self.path = path
 
     def open(self, mode='r'):
```
trnila, line 18: rb?

PR #143, opened 2021-06-14
Display max points student can get

geordi:


trnila (approved):

Kobzol (approved):
It may not be completely correct, e.g. because of tests. If there are 3 terms of a test, all of their points will be summed up. We probably need to link the tasks together somehow to recognize this.

But otherwise 👍

PR #155, opened 2022-01-07
Notifications - Improved error message when notification registration fails.

jakubhyza:
**Changes**:
- Added better error message which help strict browser users (like Brave) to fix problem with notifications.


Kobzol:
Thank you! :)

PR #158, opened 2022-01-26
Always display submit time

geordi:
Date and time of submit is sometimes needed if its long overdue.

Please, check if it works. I'm kind of bling without a working devel version :-)

### DIFF
--- a/templates/web/task_detail.html
+++ b/templates/web/task_detail.html
@@ -47,12 +47,13 @@
     -
     {{ submit.student.get_full_name }} ({{ submit.student.username }}),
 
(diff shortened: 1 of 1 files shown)


templates/web/task_detail.html
```diff
@@ -47,12 +47,13 @@ <h6>
     -
     {{ submit.student.get_full_name }} ({{ submit.student.username }}),
 
+    {{ submit.created_at|date:'d.m.Y H:i' }}
+
     {% if submit.created_at > submit.assignment.deadline %}
+    ,
```
trnila, line 53: there will be space around comma, but it is okay:)

trnila (approved):

PR #182, opened 2022-04-12
Scroll into view when clicking comment notification even if already selected

Creeperface01:
Probably a bad practice :D possibly is there any easy way to force rerender? Or should be scroll moved outside the component?

### DIFF
--- a/frontend/src/TaskDetail.svelte
+++ b/frontend/src/TaskDetail.svelte
@@ -9,7 +9,6 @@
   import {fetch} from './api.js'
   import { user } from "./global"
   import { notifications } from './notifications.js'
-  import {scroll} from "./CodeRow.svelte";
 
   export let url;
   let files = null;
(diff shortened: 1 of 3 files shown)


trnila:
thanks it looks awesome! the other way can get line element `.scrollTop` and then scroll manually with `Window.scrollTo`, but not sure if it will be easier.

frontend/src/TaskDetail.svelte
```diff
@@ -180,15 +181,17 @@
...
         const range = parts[1].split('-');
 
-        selectedRows = null;
+        selectedRows = {
+          path: parts[0],
+          from: parseInt(range[0]),
+          to: parseInt(range[1] || range[0]),
+        };
+
+        let el = document.querySelector('linecode[data-line=' + selectedRows.from + ']');
```
trnila, line 189: is it working for multiple files?

trnila: we can use this:
```
const el = document.querySelector(`table[data-path="${CSS.escape(parts[0])}"] .linecode[data-line="${CSS.escape(selectedRows.from)"]`);
```

but we need to add `data-path` to table in `SubmitSource.svelte`

trnila: we should also put this `el` inside `setTimeout` because elements dont exist yet

PR #189, opened 2022-08-23
Update OS of base Docker image to Ubuntu 22.04

Kobzol:


geordi (approved):

Kobzol:
Let's hope that upgrading `gcc` and `Python` slightly doesn't break anything :) Hopefully it shouldn't.

PR #190, opened 2022-08-23
Add support for compiling `.NET` projects

Kobzol:
It wasn't straightforward to get it working, but hopefully something like this should work for simple projects. Network access has to be enabled in order to build a standalone binary, sadly. We could get around it by building a .NET framework dependent binary, but then all additional steps (like `run`, `tests` etc.) would need to be executed inside Docker, which is annoying.

### DIFF
--- a/README.pipeline.md
+++ b/README.pipeline.md
@@ -6,10 +6,24 @@ Actions are defined in `config.yml` file in the task directory.
 
 <a href="https://github.com/mrlvsb/kelvin/blob/master/evaluator/pipelines.py">Builtin actions</a> are implemented directly in Kelvin source code, but it is also possible to run the action in <a href="https://github.com/mrlvsb/kelvin/tree/master/evaluator/images">docker container</a> with concrete compilers, so they don't need to be installed on the server.
 
+If you add a new Docker pipeline, don't forget to also modify the config validation rules
+in `frontend/src/PipelineValidation.js` (variable `rules`).
+
+### The sandbox environment limits
+Submits are evaluated in the sandboxed environment with the help of <a href="https://github.com/ioi/isolate">isolate tool</a>.
+Executing the student's submit is constrained to the <a href="https://github.com/mrlvsb/kelvin/blob/63b6ffc294e3b91d1db13453d487c773764ba4a1/evaluator/testsets.py#L122">default</a> wall clock time, memory, number of created processes etc.
+This can be overriden by the yaml configuration:
+
+```yaml
+limits:
+  fsize: 5M  # max filesize  
+  wall-time: 5 # max 5 seconds per test
+  cg-mem: 5M # max 5MB of memory usage
+```
+
 ## gcc
 Action for compiling all source codes in a submitted solution with or without a makefile.
 The errors and warnings are shown in the Result tab.
-
 
 ```yaml
 pipeline:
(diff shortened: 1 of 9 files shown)


README.pipeline.md
```diff
@@ -6,11 +6,25 @@ Actions are defined in `config.yml` file in the task directory.
 
 <a href="https://github.com/mrlvsb/kelvin/blob/master/evaluator/pipelines.py">Builtin actions</a> are implemented directly in Kelvin source code, but it is also possible to run the action in <a href="https://github.com/mrlvsb/kelvin/tree/master/evaluator/images">docker container</a> with concrete compilers, so they don't need to be installed on the server.
 
+If you add a new Docker pipeline, don't forget to also modify the config validation rules
+in `frontend/src/PipelineValidation.js` (variable `rules`).
+
+### The sandbox environment limits
+Submits are evaluated in the sandboxed environment with the help of <a href="https://github.com/ioi/isolate">isolate tool</a>.
```
trnila, line 13: I think we are not using isolate anymore? This could be removed, but limits are still configured with ulimits through docker

README.pipeline.md
```diff
@@ -19,6 +33,21 @@ pipeline:
...
+Action for compiling a .NET (core) application and producing a binary executable file as an output.
+You have to use higher limits and enable network access to make the build work, see the example
+below.
+
+```yaml
+pipeline:
+  - type: dotnet
+    output: main
+    limits:
+      network: host
```
trnila, line 46: I think we should use `bridge` instead of `host` (which is default behaviour for docker). It will create private interfaces with own subnet, otherwise pipeline will be able to start TCP/UDP servers on server address, loopback etc..

evaluator/images/dotnet/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM kelvin/base
+
+RUN apt-get update && \
+    DEBIAN_FRONTEND=noninteractive apt-get install -y \
+    dotnet6=6.0.108-0ubuntu1~22.04.1 \
+    aspnetcore-runtime-6.0=6.0.108-0ubuntu1~22.04.1 \
+    python3-pip
```
trnila, line 7: it may be good idea to remove apt caches - they are not usable after some time in following commands anyway
`&& rm -rf /var/lib/apt/lists/*`

frontend/src/PipelineValidation.js
```diff
@@ -387,6 +387,7 @@ class DockerPipeRule extends PipeRule {
             limits: [new DictRule({
               fsize: [new ValueRule(), 'Maximal size for one created file. You can use suffixes like 16M'],
               memory: [new ValueRule(), 'Maximal memory use for the container, not the process. You can use suffixes like 128M'],
+              network: [new EnumRule(["none", "host"]), '"none" to disable network access (default) or "host" to enable it'],
```
trnila, line 390: I would remove host

README.pipeline.md
Kobzol: Good idea, `bridge` works fine on my PC for this use-case.

README.pipeline.md
Kobzol: I changed the remaining mentions to Docker.

PR #199, opened 2022-10-12
Produce a better error when a test times out

Kobzol:
image

### DIFF
--- a/evaluator/images/base/Dockerfile
+++ b/evaluator/images/base/Dockerfile
@@ -7,8 +7,8 @@ RUN apt-get update && \
     g++=4:11.2.0-1ubuntu1 \
     gdb=12.0.90-0ubuntu1 \
     nasm=2.15.05-1 \
-    python3=3.10.4-0ubuntu2 \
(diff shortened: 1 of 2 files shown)


evaluator/images/base/Dockerfile
```diff
@@ -7,8 +7,8 @@ RUN apt-get update && \
     g++=4:11.2.0-1ubuntu1 \
     gdb=12.0.90-0ubuntu1 \
     nasm=2.15.05-1 \
-    python3=3.10.4-0ubuntu2 \
```
trnila, line 10: I would remove this versions entirely, otherwise we will just keep updating this file each time
(I hope locked ubuntu image should be enough to provide stable packages)

trnila (approved):

evaluator/images/base/Dockerfile
Kobzol: It's pretty annoying, yeah. But ideally I would like to keep the fixed versions, somehow by freezing `apt` :( Ignoring versions altogether is not ideal either. Anyway, let's do this in another PR.

PR #211, opened 2022-11-03
Filter out C# binary files in submit source code list

### LINKED ISSUE #210: Filterout binary files in C# submits
In `Source code` tab, filter out binary files that are in `obj` and `bin` directories in C# submits.

geordi:
Closes #210 . This has been requested by teachers.


Kobzol:
Instead of sending these directories (which can contain multiple nontrivially large files) to the client and then hiding them, it would be better if we did not send them to the client at all, or better yet, if we outright deleted them after submit.

Do we want to filter these directories globally, everywhere? This PR would have that effect basically. If yes, we can just delete/ignore them after submit.

geordi:
OK, we can delete them right away after the submit.

web/views/upload.py
```diff
@@ -90,7 +90,7 @@ def store_uploaded_file(submit: Submit, uploader: Uploader, path: str, file):
         os.unlink(target_path)
 
 
-IGNORED_FILEPATH_REGEX = re.compile(r"(__pycache__/)|(CMakeFiles/)|(\.git/)|(\.vscode/)|(\.cmake/)|(\.pyc$)")
+IGNORED_FILEPATH_REGEX = re.compile(r"(__pycache__/)|(CMakeFiles/)|(\.git/)|(\.vscode/)|(\.cmake/)|(\.pyc$)|(.*\/bin\/.*)|(.*\/obj\/.*)")
```
Kobzol, line 93:
```suggestion
IGNORED_FILEPATH_REGEX = re.compile(r"(__pycache__/)|(CMakeFiles/)|(\.git/)|(\.vscode/)|(\.cmake/)|(\.pyc$)|(bin/)|(obj/)")
```
This should be enough (for consistency with the previous entries, `__pycache__` etc. But it's just a detail.

Kobzol (approved):
LGTM.

geordi:
OK, fixed foo complicated regex :-)

trnila:
Upload filtering is done here: https://github.com/mrlvsb/kelvin/blob/master/web/views/upload.py#L15

Kobzol:
I think that it was better here to just drop the directories. They can contain various pieces of sh... :)

PR #212, opened 2022-11-11
Improve support for multiple projects in dotnet pipeline

Kobzol:
Now we try to find a `.csproj` file with `OutputType` set to `Exe`, and use its name for searching for the binary.


geordi (approved):
Tested & works.

geordi (approved):
Let's hope someone will read it.

Issue #175, opened 2022-03-10
Fix overlap of submission comments and quick buttons


Kobzol:
On the teacher page that displays user submissions, comments can overlap with quick buttons on the right:
![image](https://user-images.githubusercontent.com/4539057/157611282-a1138de3-fb77-4d4b-85cc-0c58875eb039.png)

geordi:
Looks like it does something different.

tomasmensik:
Before: https://ctrlv.link/M7AM
After: https://ctrlv.link/12cQ

PR #215, opened 2022-11-18
Comments have been overlaping with quick buttons - fix

tomasmensik:
Solves #175


frontend/src/ClassDetail.svelte
```diff
@@ -34,7 +34,7 @@
           .map(result => parseFloat(result.assigned_points))
           .reduce((acc, val) => acc + val, 0).toFixed(2);
   }
-
+  console.log(clazz)
```
geordi, line 37: Shouldn't be here.

geordi:
Looks like it does something different.

geordi:
After discussion, a new PR will be send.

PR #220, opened 2022-11-25
Fix for overlapping comments #175

tomasmensik:


tomasmensik:
Before: https://ctrlv.link/M7AM
After: https://ctrlv.link/12cQ

tomasmensik:
Before: https://ctrlv.link/8uU2
After: https://ctrlv.link/bqMh

geordi:
Please, remove the Merge commit.

trnila:
It would be better to replace absolute positioning of icons with flexbox so it is not so fragile in case we change width of page or we add new icon...

PR #224, opened 2022-12-16
Fix - Overlapping icons #175

tomasmensik:
On the teacher's page, where user submissions are displayed, comments can overlap with the quick buttons on the right.

What it looked like before: https://ctrlv.link/cLNp
What does it look like now after the fix: https://ctrlv.link/WRcl
If new icons were added in the future, it would look like this: https://ctrlv.link/LXYL

### DIFF
--- a/frontend/src/Comment.svelte
+++ b/frontend/src/Comment.svelte
@@ -36,25 +36,27 @@
     }
 </script>
 
-<div class="comment comment-{unread ? 'unread' : 'read'} {type}" on:dblclick={() => editing = can_edit}>
-  <strong>{author}</strong>: 
-  {#if !editing}
-    {#if type == 'automated'}
-      {text}
-      {#if url}
-        <a href="{url}">
-          <span class="iconify" data-icon="entypo:help"></span>
-        </a>
+  <div style="display: flex; flex-direction: row;">
+    <div class="comment comment-{unread ? 'unread' : 'read'} {type}" on:dblclick={() => editing = can_edit}>
+      <strong>{author}</strong>:
+      {#if !editing}
+        {#if type == 'automated'}
+          {text}
+          {#if url}
+            <a href="{url}">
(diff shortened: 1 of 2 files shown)


frontend/src/Comment.svelte
```diff
@@ -36,25 +36,27 @@
...
-<div class="comment comment-{unread ? 'unread' : 'read'} {type}" on:dblclick={() => editing = can_edit}>
-  <strong>{author}</strong>: 
-  {#if !editing}
-    {#if type == 'automated'}
-      {text}
-      {#if url}
-        <a href="{url}">
-          <span class="iconify" data-icon="entypo:help"></span>
-        </a>
+  <div style="display: flex; flex-direction: row;">
```
trnila, line 39: bootstrap has some helper classes like `class="d-flex flex-row"`, but i think it is okay to have it like this... At least we can easily upgrade bootstrap version - they like changing names of these helpers :(

We could also use "component style tag" which affects only elements inside this single component:
```
<style>
.container {
   display: flex;
   flex-direction: row;
}
</style>

<div class="container">
   <!-- style affects only this one class in Comment.svelte -->
</div>
```
But I would use it if we have more css properties - your solution is simpler and easier so I would keep it

trnila (approved):

frontend/src/Comment.svelte
geordi: OK, we can do it later on :-)

frontend/src/TaskDetail.svelte
```diff
@@ -215,7 +215,7 @@
     <SyncLoader />
   </div>
 {:else}
-  <div class="float-right">
+  <div class="float-right" >
```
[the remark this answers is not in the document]
geordi: Will remove in another commit :-)

tomasmensik:
Solves #175

PR #263, opened 2023-01-06
Download results per taks in CSV importable by Edison

geordi:


web/views/teacher.py
```diff
@@ -298,12 +300,38 @@ def build_score_csv(assignments, filename):
...
         return response
 
+def build_edison_task_score_csv(student_points: Dict, filename: str):
+    """
+    Build CSV file importable by Edison system.
+    Edison requires delimiter set to `;`.
+    """
+    with io.StringIO() as out:
+        for student, points in student_points.items():
+            w = csv.writer(out, delimiter=';')
```
Kobzol, line 310:
```suggestion
        w = csv.writer(out, delimiter=';')
        for student, points in student_points.items():
```
This should probably be created just once?

Kobzol (approved):
You will be using this directly by URL for now? :)

web/views/teacher.py
geordi: Thank you. I didn't catch that. And it surprises me a bit that it works.

geordi:
> 

For now, URL is enough. Will probably add a button to task edit page.

trnila (approved):

Issue #213, opened 2022-11-16
Be able to download anonymized graph generated form MOSS


geordi:
Add an option to be able to download an anonymized graph from MOSS. Currently, it's possible to download graph with logins present. This graph is present to student at the moment.

PR #265, opened 2023-01-16
Add button for downloading anonymized plagiarism graph to MOSS page

Kobzol:
Fixes: https://github.com/mrlvsb/kelvin/issues/213


geordi (approved):
Works for me.

Just a note. We'd need to have a LUT to notify student what number he/she is in the anonymized version, so we don't have to compare it by eye. But it works great anyway.

Issue #178, opened 2022-03-10
Display total points per assigned task in class table


Kobzol:
On the teacher dashboard, there is a table per class that displays individual students, their points for each assigned task, and the total amount of points per student. It would be nice to add a total amount of points per assigned task on a row at the bottom of the table.

Kobzol:
Implemented in https://github.com/mrlvsb/kelvin/pull/282.

PR #282, opened 2023-02-24
Total points per assigned task #178

tomasmensik:
Solves: #178 
I've added the total points per assignment in the class of students. 

Here is what it looked like before: [Click here](https://ctrlv.link/jlOy) (There is nothing in the bottom row of the table)

Here is what it looks like now: [Click here](https://ctrlv.link/Vzt4) (There is sum of points per assignment in the bottom row of the table)

### DIFF
--- a/frontend/src/ClassDetail.svelte
+++ b/frontend/src/ClassDetail.svelte
@@ -35,6 +35,18 @@
           .reduce((acc, val) => acc + val, 0).toFixed(2);
   }
 
+  function taskPoints(clazz, k) {
+  let assignmentPoints = 0;
+
+  for (let i = 0; i < clazz.students.length; i++) {
+    const student = clazz.students[i];
+      if (!isNaN(clazz.assignments[k].students[student.username].assigned_points)) {
+        assignmentPoints += clazz.assignments[k].students[student.username].assigned_points;
+      }
+  }
+  return assignmentPoints;
+}
+
   let showFullTaskNames = localStorageStore('classDetail/showFullTaskNames', false);
   let showSummary = false;
 </script>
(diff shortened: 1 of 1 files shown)


frontend/src/ClassDetail.svelte
```diff
@@ -35,6 +35,18 @@
           .reduce((acc, val) => acc + val, 0).toFixed(2);
   }
 
+  function taskPoints(clazz, k) {
```
Kobzol, line 38: Please fix the formatting of this function.

frontend/src/ClassDetail.svelte
```diff
@@ -35,6 +35,18 @@
...
+  let assignmentPoints = 0;
+
+  for (let i = 0; i < clazz.students.length; i++) {
+    const student = clazz.students[i];
+      if (!isNaN(clazz.assignments[k].students[student.username].assigned_points)) {
```
Kobzol, line 43: Sometimes we assign negative points. I think that it would be better to ignore these in the summary and treat any negative points as zero. You can add `Math.max(0, ... assigned_points)` to the running total.

frontend/src/ClassDetail.svelte
```diff
@@ -229,6 +241,16 @@ tr:hover td:first-of-type {
               <td>{studentPoints(clazz, student)}</td>
             </tr>
             {/each}
+            <tr>
+              <td>&nbsp;</td>
```
Kobzol, line 245: Why are the non-breaking spaces here? The text in the columns should already be center aligned.

Kobzol (commented):
Thank you! The screenshot looks nice. I left a few comments.

frontend/src/ClassDetail.svelte
tomasmensik: Fixed in the next commit called "Updated totalPointsPerAssignedTask".

frontend/src/ClassDetail.svelte
tomasmensik: Fixed in the next commit called "Updated totalPointsPerAssignedTask".

frontend/src/ClassDetail.svelte
tomasmensik: There has to be 2 <td> tags because there are 2 columns that serve for login code and student's name. But I agree that there are extra non-breaking spaces. I changed the non-breaking spaces just for an empty <td> tag. It is the same outcome. 

Fixed in the next commit called "Updated totalPointsPerAssignedTask".

Kobzol (approved):
Looks great, thank you!

Issue #193, opened 2022-09-23
Integrate INBUS


geordi:
Integrate information available in VSB's INBUS system.

geordi:
According to latest info, the API we need will be available in future...

geordi:
As of now, needed INBUS APIs are ready.

PR #283, opened 2023-02-24
INBUS integration to get user info

geordi:
Partially solves #193 .

This if the first INBUS API call we can make. More to come.

### DIFF
--- a/common/inbus.py
+++ b/common/inbus.py
@@ -51,12 +51,12 @@ def request_new_token() -> Dict:
 
 
 def is_response_ok_or_new_token_(response: requests.Response) -> bool:
-    if response.status_code == 404:
+    if response.status_code == requests.codes.OK:
+        return True
+    elif response.status_code == requests.codes.UNAUTHORIZED:
         token = request_new_token()
         set_token_to_cache(token)
         return False
-    elif response.status_code == 200:
-        return True
     else:
         return False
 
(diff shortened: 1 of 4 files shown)


common/inbus.py
```diff
@@ -0,0 +1,105 @@
...
+    return token
+
+
+def request_new_token() -> Dict:
+    token = authenticate()
+    return token
+
+
+def is_response_ok_or_new_token_(response: requests.Response) -> bool:
+    if response.status_code == 404:
```
Kobzol, line 54: It returns 404 when authentication is not correct? Nice.. :)

common/inbus.py
```diff
@@ -0,0 +1,105 @@
...
+    if not person_resp:
+        return None
+    person_json = person_resp.json()
+
+    return person_json
+
+
+# Kelvin's interface
+
+def search_user(login: str) -> Dict:
```
Kobzol, line 98: Could this return a dataclass instead? We pass this data to the constructor of `User` through `**`, that is quite brittle :)

Kobzol (approved):

common/inbus.py
geordi: Sorry, my bad. It's `401` actually. I'll fix it in next commit.

Kobzol (approved):
LGTM

PR #285, opened 2023-03-03
Use INBUS API to fill user info instead of LDAP

geordi:
INBUS API response may have missing fields. So at this point simeple dict.get with default is used. Maybe we can use something like Dacite and default values in dataclass.


Kobzol (approved):
It would be nice if `inbus` was a bit more separated from Django, so that we could actually write tests for it easily :) It's a pretty important part of Kelvin.

Kobzol (approved):
LGTM

PR #286, opened 2023-03-04
Allow checking multiple tasks together with MOSS

Kobzol:
This PR adds the ability to link multiple tasks together via a `plagiarism_key` string. All tasks with the same plagiarism key from the same subject will be checked for plagiarism together.

As a follow-up, it would be really nice to add autocomplete for the plagiarism key input field, pre-filled with existing plagiarism keys.

Best reviewed commit by commit.


geordi (approved):

geordi (approved):
Works for me :-)

PR #289, opened 2023-03-09
Disabling `draggable` attribute

Astra3:
Disabling `draggable` attribute on notification icon so that the icon can't be dragged and submitted when released, like happened to me. The submitted work looks like this:
Screenshot_20230309_094425

### DIFF
--- a/frontend/src/Notifications.svelte
+++ b/frontend/src/Notifications.svelte
@@ -26,7 +26,7 @@ async function openNotification(notification) {
 {#if $notifications}
 <span style="position: relative" use:clickOutside on:click_outside={() => show = false}>
     <span on:click={() => show = !show} style="cursor: pointer">
-      <img src="/static/notify_icon.png" style="height: 15px;" />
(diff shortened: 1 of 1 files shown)


Astra3:
It may be possible to read what is being dragged and not allow the drag based on that, for example with filename matching (that could disallow files of the same name being uploaded) or some kind of hashing.

Or realistically, to my knowledge only `img` elements can be dragged and dropped as files (SVG icons can't be dragged). So just putting `draggable="false"` on all of them might work. The only `img` elements on the site seem to be the notification bell and [files uploaded by students and teachers](https://github.com/mrlvsb/kelvin/blob/f4dd5b80cda99c87554af0e3726c7b514a11a4cf/frontend/src/TaskDetail.svelte#L266) shown in task results (yes I can just drag the uploaded bell and drop it to upload it again, just noticed) or in task descriptions.

Kobzol (approved):
Ok, then let's keep it like this. Thanks!

PR #293, opened 2023-03-13
Adding dark mode

Astra3:
Added simple dark mode by removing `bootstrap` as node dependency and using CDN to fetch [bootstrap-dark](https://github.com/vinorodrigues/bootstrap-dark). The ideal solution how to tackle this problem would be to migrate to Bootstrap 5.3.0, but this might work great temporarily.

Marking as draft as issues with icons and stuff like syntax highlight are expected and the current solution is a bit dirty too.


Astra3:
This PR is to become obsolete by migrating to Boostrap 5 (see #296) which is without a doubt a better solution.

PR #294, opened 2023-03-16
Show semester(s) where a linked task was assigned in MOSS page

Kobzol:
So that we have a better idea of where do the tasks come from.


geordi (approved):
LGTM

PR #296, opened 2023-03-21
Move from Boostrap 4.6 to 5.3 and add dark mode

Kobzol:
This should enable us to use Bootstrap's dark mode relatively easily (we will still need to perform some UI changes though, some colors are hardcoded).

@Astra3 has implemented theme switcher and has fixed many places of the website. Thank you very much!


geordi (approved):
Let's see :-)

PR #306, opened 2023-04-03
Fix displaying of MOSS sources

Kobzol:
Before, MOSS result matches were automatically filtered and the match list was indexed by match ID. However, if some entries were filtered (which happened almost all the time), indexing by ID would yield incorrect results. Now matches are looked-up using their ID instead of using it directly as a list index.

Additionally, MOSS results can now be displayed for results that do not pass through the currently active filter. Adding the flag is dirty, but the `MossResult` API would need to be refactored completely to solve this easily in a different way.

### DIFF
--- a/web/views/moss.py
+++ b/web/views/moss.py
@@ -14,7 +14,8 @@ from django.views.decorators.clickjacking import xframe_options_sameorigin
 from notifications.models import Notification
 
 from common.models import Semester, Submit, Task, current_semester
-from common.moss import PlagiarismMatch, enqueue_moss_check, get_linked_tasks, get_match_local_dir, \
+from common.moss import PlagiarismMatch, enqueue_moss_check, get_linked_tasks, \
+    get_match_local_dir, \
     moss_delete_job_from_cache, \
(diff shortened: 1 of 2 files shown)


geordi:
Will look at it tomorrow.

web/views/moss.py
```diff
@@ -114,7 +115,7 @@ def task_moss_check(request, task_id):
             assignment__task_id=task.id,
             created_at__gt=res.started_at
         ).count()
-        matches = enrich_matches(res.matches, request.user, task)
+        matches = enrich_matches(rse.matches, request.user, task)
```
geordi, line 118: Looks like a typo. Should be `res`.

geordi (changes requested):
Just fix the typo.

geordi:
At this point, somehow I don't see matches in the table but I see them in the graph. But this behavior I experience on running system without this PR.

web/views/moss.py
Kobzol: Thanks!

Issue #184, opened 2022-05-02
Allow grouping task points into categories


Kobzol:
It would be nice to categorize tasks into custom categories (lessons, tests, projects, etc.). The table for teachers on the homepage should then show the points awarded to individual students amongst the categories.

The categories should be set in the task edit page. They can be hardcoded for simplicity first.

Kobzol:
Duplicate of https://github.com/mrlvsb/kelvin/issues/578.

PR #307, opened 2023-04-21
Filter #184

tomasmensik:
ISSUE #184 
"It would be nice to categorize tasks into custom categories (lessons, tests, projects, etc.). The table for teachers on the homepage should then show the points awarded to individual students amongst the categories.

The categories should be set in the task edit page. They can be hardcoded for simplicity first."

----------------------------------------------------------------------------------------------------------------------------------------------------------------------------

I didn't really fully understand the "homepage thing" of the issue but what I did understand was the categorization. I made a simple filter in the admin tab "Tasks". So, now you can filter through the tasks by either buttons you wanted (lessons, tests, projects) or by simply typing a name of the task into the search bar.

I didn't do the homepage points. But the filter/categorization works just fine.

**Before:**  https://ctrlv.link/LKrr
**After - default:**  https://ctrlv.link/aqb6
**After - button clicked:**  https://ctrlv.link/64rd
**After - search:**  https://ctrlv.link/HrhV

### DIFF
--- a/templates/web/all_tasks.html
+++ b/templates/web/all_tasks.html
@@ -1,29 +1,87 @@
 {% extends 'web/layout.html' %}
 
 {% block content %}
-{% for subj in subjects %}
-  <a href="{% url 'tasks' subj.abbr %}">{{ subj.abbr }}</a>
-{% endfor %}
 
-  <table class="table table-striped table-hover table-sm">
-    <thead>
-      <tr>
-        <th>Name</th>
-        <th class="text-right"></th>
-      </tr>
-    </thead>
+<table class="table table-hover table-sm table-borderless">
+  <thead>
+    <tr>
+      <th class="text-left" colspan="4">Filter by</th>
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
+      <th class="text-left btn-group">
+        <button title="This button filters all the tasks by the given subject." id="toggle-button"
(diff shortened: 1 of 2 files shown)


Kobzol:
Hi! The UI looks great, but before we start filtering or displaying tasks by their category, we should first implement assigning of categories to individual tasks, to avoid heuristics based on the task name.

So I would suggest to first implement a different PR, that will:
1) Add category column to Task (it probably makes more sense to be in Task and not AssignedTask). I would use a (nullable) string for it, to allow arbitrary values eventually and to make it simple.
2) Add a select box to TaskEdit page where teachers can select the task category. The default category should be lesson, and if it's NULL in the DB, it should also be treated as a lesson. I would use hard-coded categories "Lesson", "Test" and "Project" in the select box for now.

Once we have this data in the DB, we can use it to filter tasks by categories in the Tasks page, and also to split different task categories in the Teacher homepage.

tomasmensik:
Hey, unfortunately, I don't have access to the source code anymore. But those who want to finish it can.

The UI is fine, as you say, and it can be left there, and all that's left is to better optimize my code to your requirements.

Kobzol:
If you want to continue working on this PR, you can just download the source code from GitHub :)
```
$ git clone https://github.com/tomasmensik/kelvin
$ git checkout filter
```

Issue #149, opened 2021-09-16
Remove of student from class


geordi:
It would be nice to be able to remove a student from a class.

PR #308, opened 2023-04-24
Removing students from class

HansS04:
Solves ISSUE #149 

I have created a side button with a trash can icon which when clicked will create another column of icons. Click the trash can button on a row with a specific student to remove the link between that student and the class. And that removes the user from the class.

There are also treated conditions, such as "is_staff" etc.

Once everything goes well, a message will pop up saying that the user has been successfully removed from the class.


Kobzol:
Hi! :) Could you please remove the extraneous changes (the `.idea` directory and the formatting changes) and keep only the changes related to the PR? Because of the committed `.idea` I can't even check out the code :(

PR #324, opened 2023-07-01
Adding theme picker and updating navbar

Astra3:
I've added a theme picker to navbar and updated the navbar to account for dark theme. The page also now responds to CSS `prefers-color-scheme` media query. The navbar's collapsing feature also now uses bootstrap's implementation instead of a custom one (it seems smarter that way).

There are still some issues in the UI with white backgrounds, namely in code snippets and in some parts of the teacher's UI, so it's not ready to be merged into master. Also there can be some contrast issues with hard-coded colors, but I have yet to found any.


Astra3:
The support for Bootstrap 5 and dark mode should be complete. Only `diff2html` is forced to light mode due to missing support for dark mode from the library.

Kobzol:
This is great, thank you! :heart:

Kobzol (approved):

PR #334, opened 2023-07-12
Allow configuring sentry URL

Kobzol:
So that it doesn't need to be applied using git working dir modifications.


geordi (approved):
All set :)

Issue #152, opened 2021-09-24
`/import` fails


geordi:
`/import` fails when there's a teacher without diacritics in `Users` but in schedule diacritics is present.

Problem is probably somewhere here [bulk_import.py#L87](https://github.com/mrlvsb/kelvin/blob/master/common/bulk_import.py#L87).

geordi:
OK, we also need a safe way to import users (teachers) with diacritics. So far, if I wanted to create a user (teacher), I added him as a student to a class, which creates a new user __but without diacritics__. Then I remove the user from class and set teacher group in admin.

trnila:
We can create teachers individually in django admin - is it enough?
I think EDISON export has only teacher name, but without login...

geordi:
Will at the INBUS interface and see what's available there.

geordi:
INBUS provides names with diacritics. Also, import of classes can be made from INBUS. This is more broadly discussed in #193.

PR #341, opened 2023-07-18
Inbus import

geordi:
Adds importing classes from INBUS so we don't need to save a schedule HTML from Edison. Still, some rough edges exist, but it's a good start.

Closes: #152
Addresses: #193


common/inbus/dto.py
```diff
@@ -0,0 +1,160 @@
...
+
+
+@serde.serde
+@dataclass
+class StudyRelation:
+    """
+    Relation of student and university.
+    """
+
+    studyRelationId: int # ID studijního vztahu
```
Kobzol, line 124: Maybe we should list only the attributes that we care about? To make the code more readable and also to limit the number of fields that will break deserialization once they remove them :)

api/views.py
```diff
@@ -557,3 +578,48 @@ def transfer_students(request):
...
+    # TODO: When upgrading, see: https://docs.djangoproject.com/en/4.2/ref/request-response/#httpresponse-objects
+    # for a way to set content type.
+    return HttpResponse(semesters_json,
+                        content_type="application/json")
+
+
+
+
+@user_passes_test(is_teacher)
+def import_activities(request):
```
Kobzol, line 598: `@require_http_methods(["POST"])`

api/views.py
```diff
@@ -557,3 +578,48 @@ def transfer_students(request):
...
+    res = {}
+    if request.method != 'POST':
+        return HttpResponseBadRequest()
+
+    post = json.loads(request.body.decode('utf-8'))
```
Kobzol, line 603: Duplicated line.

frontend/src/ImportFromInbus.svelte
```diff
@@ -0,0 +1,197 @@
...
+    {:else}
+        Import
+    {/if}
+</button>
+
+
+{#if result}
+    <table class="table table-sm table-hover table-striped">
+        <tbody>
+        {#each result.users as item}
```
Kobzol, line 187: I got this error here after importing classes:
```
index.mjs:2374 Uncaught (in promise) Error: {#each} only iterates over array-like objects.
    at validate_each_argument (index.mjs:2374:15)
    at create_if_block$a (ImportFromInbus.svelte:187:28)
    at Object.update [as p] (ImportFromInbus.svelte:184:12)
    at update (index.mjs:1347:36)
    at flush (index.mjs:1307:17)
```
because the import has failed:
```
"Traceback (most recent call last):\n  File \"/projects/vsb/upr/kelvin/common/bulk_import.py\", line 191, in run_inbus\n    class_in_db[c] = Class.objects.get(code=c, semester=semester, subject=subject)\n  File \"/projects/vsb/upr/kelvin/venv/lib/python3.10/site-packages/django/db/models/manager.py\", line 85, in manager_method\n    return getattr(self.get_queryset(), name)(*args, **kwargs)\n  File \"/projects/vsb/upr/kelvin/venv/lib/python3.10/site-packages/django/db/models/query.py\", line 429, in get\n    raise self.model.DoesNotExist(\ncommon.models.Class.DoesNotExist: Class matching query does not exist.\n\nDuring handling of the above exception, another exception occurred:\n\nTraceback (most recent call last):\n  File \"/projects/vsb/upr/kelvin/api/views.py\", line 615, in import_activities\n    res['users'] = list(BulkImport().run_inbus(activities, subject, semester))\n  File \"/projects/vsb/upr/kelvin/common/bulk_import.py\", line 213, in run_inbus\n    class_in_db[c].save()\n  File \"/projects/vsb/upr/kelvin/venv/lib/python3.10/site-packages/django/db/models/base.py\", line 700, in save\n    raise ValueError(\nValueError: save() prohibited to prevent data loss due to unsaved related object 'teacher'.\n"
```

Kobzol (commented):
This is very cool!

api/views.py
geordi: Oh, I didn't know that decorator. I found there's also `@require_POST()`, so I'll use this one directly.

api/views.py
geordi: Thanks!

common/inbus/dto.py
geordi: Well, you're right. I think this is also used by Edison, so there's low probability of changing/breaking. Also, I'd like to separate this INBUS implementation into a completely different package and use it for broader use cases and this is "kind of" testbed :)

common/inbus/inbus.py
```diff
@@ -0,0 +1,71 @@
...
+    offset = 0
+    while True:
+        url = urllib.parse.urljoin(config.INBUS_SERVICE_EDISON_URL, 'edu/subjectVersions')
+        subject_versions_resp = utils.inbus_request(url, {'departmentId': department_id, 'offset': offset, 'limit': results_per_page})
+
+        subject_versions_json = subject_versions_resp.json()
+
+        for subject_version_json in subject_versions_json:
+            subject_json = subject_version_json['subject']
+            subject_guarantee = serde.from_dict(dto.Person, subject_version_json['guarantee'])
```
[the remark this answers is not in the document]
geordi: Good point. There's `SubjectVersion`'s `guarantee` and also `Subject`'s `guarantee`. It's kind of complicated and a bit overkill, but in the future, we'll probably have to use it. Anyway, I'll fix it now.

frontend/src/ImportFromInbus.svelte
geordi: Tnaks for the report. I'm able to reproduce a failed import with yet another error :)

geordi: OK. `User` was not `save`d.

PR #343, opened 2023-08-11
Add `cargo` pipeline

Kobzol:
Can be used with `cargo build`, `cargo test`, `cargo run`, `cargo clippy` etc. It also prepares an executable (if available) under the filepath `main`, to be used with follow-up `run` pipelines.


geordi (approved):
In Rust we trust :-)

Issue #348, opened 2023-09-20
Remove 'Add class' icon for non staff users


geordi:
When clicked it shows 403 Forbidden, which is correct but this link shouldn't be provided anyway.

Kobzol:
Fixed by https://github.com/mrlvsb/kelvin/pull/349

PR #349, opened 2023-09-22
Remove + icon for non-staff users (#348)

geordi:


frontend/src/ClassList.svelte
```diff
@@ -88,9 +88,11 @@
           <span class="iconify" data-icon="mdi:calendar-import"></span>
         </a>
 
+        {#if $user.is_staff}
```
Kobzol, line 91: What does `$` do here? :)

geordi: I have no idea, but without it, it doesn't work at all :)

geordi: Well, IMO, it forces `user` to load actually.

Kobzol: Ok :)

Kobzol (approved):

Issue #355, opened 2023-09-27
Sort class list in filter drop down menu


geordi:
It would be nice to have list of classes sorted by time as it it right now in the main page.

geordi:
Solved in #361.

Kobzol:
It would be nice to display the date and time in the select list :)

PR #361, opened 2023-10-04
Sort classes in class filter by date/time (#355)

geordi:


Kobzol (commented):
It's Kelvin, so it doesn't really matter, but sorting items in an API endpoint for a specific UI usage is pretty dirty :) Not to mention the `setattr`. Why can't it be sorted like this:
```python
classes = sorted(classes, key=lambda klass: (day_mapper[klass.day], klass.time)) 
```

geordi:
I agree. Your solution work. Thank you!

PR #363, opened 2023-10-06
Display message if a newline is missing at eof

fpeterek:


evaluator/pipelines.py
```diff
@@ -182,6 +182,36 @@ def run(self, evaluation):
         }
 
 
+def with_nl_message(diff: str):
```
Kobzol, line 185: Could you please add some short description of what this function does and how does it modify the diff output? Also, it should document that it needs the `diff -u` flag, otherwise it won't work.

Kobzol (approved):
Great, thank you!

PR #367, opened 2023-10-12
Support for Java language 

kozusznik:
Docker image for Java evaluator.

### DIFF
--- a/evaluator/images/java/entry.py
+++ b/evaluator/images/java/entry.py
@@ -165,7 +165,17 @@ result = build_java_project(run_tests)
 
 with open("result.html", "w") as out:
     stdout = bleach.clean(result.output.strip()).replace("\n", "<br />")
-    out.write(f'<pre>{stdout}</pre>')
+    stdout = f"<pre><code>{stdout}</code></pre>"
+    if result.success:
+        out.write(f"""<details>
+<summary>Stdout</summary>
+
+{stdout}
+
+</details>
+""")
+    else:
+        out.write(stdout)
 
 with open('piperesult.json', 'w') as out:
     json.dump({"comments": result.comments, "tests": result.tests}, out, indent=4, sort_keys=True)
--- a/evaluator/images/java/entry.py
+++ b/evaluator/images/java/entry.py
@@ -1,4 +1,4 @@
-#!/usr/local/bin/python3
+#!/usr/bin/python3
 import dataclasses
 import os
 import subprocess
@@ -30,11 +30,11 @@ class BuildResult:
 
 def parse_tests_report(path):
     try:
+        tests = []
         for path in glob.glob(os.path.join(path, "*.xml")):
             tree = ET.parse(path)
             os.unlink(path)
 
-            tests = []
             for testcase in tree.findall('.//testcase'):
                 success = True
                 message = ""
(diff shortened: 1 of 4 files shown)


Kobzol:
Hi, thanks, this looks good! Do you perhaps have some Java test project with the intended layout (e.g. a set of Java files, Maven configuration file, etc.), so that I could test it locally?

kozusznik:
> Hi, thanks, this looks good! Do you perhaps have some Java test project with the intended layout (e.g. a set of Java files, Maven configuration file, etc.), so that I could test it locally?

Hi, thank you. You can use this [example](https://github.com/kozusznik/test-java/blob/master/README.md).

evaluator/images/java/entry.py
```diff
@@ -0,0 +1,179 @@
+#!/usr/local/bin/python3
```
Kobzol, line 1:
```suggestion
#!/usr/bin/python3
```

evaluator/images/java/entry.py
```diff
@@ -0,0 +1,179 @@
...
+        )
+
+
+def parse_tests_report(path):
+    try:
+        for path in glob.glob(os.path.join(path, "*.xml")):
+            tree = ET.parse(path)
+            os.unlink(path)
+
+            tests = []
```
Kobzol, line 37: This will return only tests from the last `xml` file. Either `tests = []` shold be defined before the cycle (if tests from all `.xml` files should be returned), or `return tests` should be indented one level to the right (if only the tests from the first `.xml` file should be returned).

evaluator/images/java/entry.py
```diff
@@ -0,0 +1,179 @@
...
+    for java_file in find_java_files(directory):
+        if is_executable_java_file(java_file):
+            executable_classes.append(get_qualified_class_name(java_file, directory))
+    return executable_classes
+
+
+def build_java_project(run_tests: bool) -> BuildResult:
+    output_dir = "output"
+    paths = os.listdir(os.getcwd())
+    pom_xml = [p for p in paths if Path(p).suffix == "pom.xml"]
```
Kobzol, line 88:
```suggestion
    paths = os.listdir(os.getcwd())
```
These two variables were unused.

Kobzol (changes requested):
Apart from the mentioned nits, I have also pushed two changes that we need to make to properly support the Java pipeline.

I haven't tested tests, but build of the linked example project looks fine. The build log is quite… verbose, so it might be nice to hide it by default in a `<details>` HTML tag, like is being done for Rust ([like this](https://github.com/mrlvsb/kelvin/blob/f37604179a603f6be3cf5a31611df6f88f942ede/evaluator/images/cargo/entry.py#L190)). But it's just a nice-to-have thing.

evaluator/images/java/entry.py
Kobzol: Otherwise `python3` is not found.

evaluator/images/java/entry.py
kozusznik: Thank you for you review. I will reflect your notes.

kozusznik: Actuallu, the variable paths is alson not used. I will remove it.

kozusznik:
Thank you again for the review. I addresed all comments.

kozusznik:
Thank you again for the review. I addresed all comments.

Kobzol (approved):
Thanks! I pushed one additional fix, since the code should have newlines and not `<br />`, since it's displayed in `<pre>`.

PR #370, opened 2023-10-17
INBUS: Change caching of listing subject versions

geordi:
This enables caching of INBUS request to get all subject versions system wide, not only through web API.


geordi:
OK, will do it differently.

PR #371, opened 2023-10-18
Kelvin vs Edison schedule difference script

geordi:
The script prints all classes and students that should be attend a different class.

### DIFF
--- a/common/management/commands/kelvin_edison_difference.py
+++ b/common/management/commands/kelvin_edison_difference.py
@@ -1,0 +1,55 @@
+from django.core.management.base import BaseCommand
+
+from django.core.cache import cache
+
+from common.models import Class, Subject
+from common.inbus import inbus
+
+
+class Command(BaseCommand):
+    def add_arguments(self, parser):
+        parser.add_argument('subject')
+
+    def handle(self, *args, **opts):
+        subject = Subject.objects.get(abbr__icontains=opts['subject'])
+        subject_versions = cache.get('subject_versions')
+
+        if not subject_versions:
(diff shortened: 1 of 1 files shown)


common/management/commands/kelvin_edison_difference.py
```diff
@@ -0,0 +1,55 @@
...
+                                print(f"\tIN K {k}")
+                                other_ca = mapping.get(k, None)
+                                if other_ca:
+                                    print(f"\t\tShould be in {other_ca.code()} by {other_ca.teacherLogins}")
+                        for i in sorted(students_inbus):
+                            if i not in students_class:
+                                print(f"\tIN E {i}")
+
+                    except Class.DoesNotExist:
+                        pass
```
Kobzol, line 54: Maybe we could also print that a class does not exist?

common/management/commands/kelvin_edison_difference.py
```diff
@@ -0,0 +1,55 @@
...
+                    try:
+                        clazz = Class.objects.get(code=code)
+                        print(f"{ca.code()} by {ca.teacherLogins}")
+
+                        students_class = [u.username.upper() for u in clazz.students.all()]
+
+                        for k in sorted(students_class):
+                            if k not in students_inbus:
+                                print(f"\tIN K {k}")
+                                other_ca = mapping.get(k, None)
```
Kobzol, line 46:
```suggestion
                                other_ca = mapping.get(k)
```
nit: redundant

Kobzol (approved):
Very cool! Should be incredibly useful :)

PR #374, opened 2023-10-21
Include runner script in tests archive

DanielKrasny:
To simplify local testing, a bash script was created. It compares the executable's standard output with `stdout` file (and similarly for `stderr`). On differences a `git diff` output is saved. It is now included with every downloaded archive of all tests. Due to many different types of tasks `args` are bundled as a separate file and if it is present, then it is passed to executable.

### DIFF
--- a/web/views/assets/run-tests.sh
+++ b/web/views/assets/run-tests.sh
@@ -1,0 +1,42 @@
+#!/bin/bash
+
+TABLE_OUTPUT=""
+BINARY=${1:-./main}
+
+if [ -f ${BINARY} ]
+then
+  RESULT=""
+  for dir in */
+  do
+    ARGS=""
+    if [ -f ${dir}args ]
+    then
+      ARGS=$(cat ${dir}args)
+    fi
+    ${BINARY} ${ARGS} < ${dir}stdin > ${dir}real_stdout 2> ${dir}real_stderr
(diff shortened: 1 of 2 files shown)


Kobzol:
This is great, thank you!

I think that we can make this even better. Since we know the exact test structure, we can render a bash script that will be specific to the test set. It would also make the argument handling much easier, as we would simply render the args directly into the script. So we could generate something like this:
```bash
$ ${BINARY} foo bar baz 1 2 < test1/input > out
$ test $? 1 # check exit code with some bash magic

$ ${BINARY} 3 4 < test2/input > out
etc.
```

Let me know if you're interested in implementing this. If yes, we can discuss it further, if not, I'll merge this and implement it some time later :)

DanielKrasny:
With the power of bash functions I have reimplemented the script. When creating an archive with tests, every test is included in the final script by just calling:
```
execute_test 'test1' $'foo bar \'baz\' 1 2' '1'
```

So the `args` file is not included again and exit code is verifying too. I could simplify the script to allow appending these calls, but I ended up with replacing a comment called `# --kelvin-generate--`.

Kobzol:
I was thinking more something like creating a Django template and rendering the bash script directly, without needing to do these hacky replacements. But this is already much better than before :) Thank you. I will try it out locally tomorrow.

Kobzol (approved):
Thank you! I have implemented a few QoL improvements. The script is really useful!

PR #375, opened 2023-10-22
Ask before uploading dropped solution

DanielKrasny:
When dropping files to browser, a question modal is shown to make sure user really wants to submit these files. Modal also contains list of dropped files.
Example:
image
However, in some cases, I received this file list and I think I can't impact that behavior since these uploads will fail.
image
The modal component was retrieved and modified from https://eternaldev.com/blog/using-bootstrap-modal-in-svelte.

### DIFF
--- a/frontend/src/Modal.svelte
+++ b/frontend/src/Modal.svelte
@@ -1,0 +1,46 @@
+<script>
+import {fade, fly} from "svelte/transition";
+import {quintOut} from "svelte/easing";
+
+export let open = false;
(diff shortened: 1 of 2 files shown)


geordi:
Nice addition, thank you!

We'll look at it ASAP.

Kobzol (approved):
This is incredible, thank you so much! In order to save time for me a bit, I implemented a few changes, I hope you don't mind. Thanks!

PR #380, opened 2023-10-28
Add background to code in comments

DanielKrasny:
As requested by @fpeterek. Applies only for dark mode.
Example:
image

### DIFF
--- a/frontend/src/global.scss
+++ b/frontend/src/global.scss
@@ -9,6 +9,13 @@ html[data-bs-theme="dark"] {
   @include meta.load-css('node_modules/flatpickr/dist/themes/dark.css');
   @include meta.load-css('node_modules/codemirror/theme/dracula.css');
 
+  .comment code {
+    display: inline-block;
+    background-color: var(--bs-card-bg);
+    border: var(--bs-card-border-width) solid var(--bs-card-border-color);
+    border-radius: var(--bs-card-border-radius);
+  }
+
   .code-diff {
     .d2h-ins {
       background-color: #748a74;


Kobzol:
The comment colors are set here: https://github.com/mrlvsb/kelvin/blob/85f262b287f7fd63a98610911cea0fb6bc42ac0d/frontend/src/CodeRow.svelte#L72
We should thus modify this CSS, instead of hacking the global CSS file :)

I tried to experiment a bit and came up with this:
```css
  :global(html[data-bs-theme="dark"] .comment) {
      --bs-code-color: #0021FF;
      color: #000000;
  }
  :global(html[data-bs-theme="dark"] .comment.teacher) {
      background: #FFFF2E;
  }
  :global(html[data-bs-theme="dark"] .comment.teacher.comment-read) {
      background: #C0C035;
  }
  :global(html[data-bs-theme="dark"] .comment.student) {
      background: #4EFF0E;
  }
  :global(html[data-bs-theme="dark"] .comment.student.comment-read) {
      background: #66EC36;
  }
  :global(html[data-bs-theme="dark"] .comment.automated) {
      background: #1B96FF;
  }
```

It's hopefully better than before, but still not *super nice*. If you can find colours with a better contrast, I'm all in :) I'd just like to keep similar colors as with the light scheme, so yellow for teachers, green for students and blue for automation comments.
image

DanielKrasny:
I wanted to address syntax highlighted comments, so these could not be fixed with a single color in my opinion. However, I also agree that proposed colors look much better than before. Personally, I think we could move it from global styles, add your modifications and set the background only for comment parts containing something like:
~~~
```c
void moveTurtle(Turtle* const turtle) { ... }
```
~~~
Defining colors for multiple languages might get complicated :)

PR #381, opened 2023-10-28
ClassDetail: Treat null as NaN

### LINKED ISSUE #379: Points statistics not correct
I thing, point stats is not displaying correct data, see screenshot below:

points_stats

DanielKrasny:
Closes #379.

### DIFF
--- a/frontend/src/ClassDetail.svelte
+++ b/frontend/src/ClassDetail.svelte
@@ -60,8 +60,9 @@
 
       for (let i = 0; i < clazz.students.length; i++) {
           const student = clazz.students[i];
-          if (!isNaN(parseFloat(clazz.assignments[assignment_index].students[student.username].assigned_points))) {
-              assignmentPoints += Math.max(0, clazz.assignments[assignment_index].students[student.username].assigned_points);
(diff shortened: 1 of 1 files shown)


frontend/src/ClassDetail.svelte
```diff
@@ -60,7 +60,7 @@
 
       for (let i = 0; i < clazz.students.length; i++) {
           const student = clazz.students[i];
-          if (!isNaN(clazz.assignments[assignment_index].students[student.username].assigned_points)) {
+          if (!isNaN(parseFloat(clazz.assignments[assignment_index].students[student.username].assigned_points))) {
```
Kobzol, line 63: This is a bit hacky. Could we do something like this instead?
```suggestion
          const points = clazz.assignments[assignment_index].students[student.username].assigned_points;
          if (!isNaN(points) && points !== null) {
```

Kobzol (commented):
Thanks for the quick fix :)

frontend/src/ClassDetail.svelte
DanielKrasny: Of course we could :) I was just afraid of creating a variable for that.

Kobzol (approved):
Thank you! :heart:

PR #382, opened 2023-10-28
Improve path handling in `run-tests.sh` script

DanielKrasny:
The file existence condition was useless because of running `realpath` before checking path existence.
Script paths were improved to allow running outside of a current directory.
A check for git existence was added. (Why don't just everyone use git though?)
Since we have some macOS users with really weird default terminal, the ANSI escape sequences were replaced with `tput` command.

### DIFF
--- a/web/views/assets/run-tests.sh
+++ b/web/views/assets/run-tests.sh
@@ -6,8 +6,6 @@
 TABLE_OUTPUT=""
 BINARY=${1:-./main}
 SCRIPT_PATH=$(dirname $(realpath $0))
-command -v git &> /dev/null
-HAS_GIT=$?
 
 execute_test () {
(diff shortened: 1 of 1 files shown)


Kobzol:
I can't really do a lot of magic with bash, so I'm gonna trust your instincts on this being better :) Btw without git we could just use a normal `diff` command, but that might not always be available, I suppose.

DanielKrasny:
Good note, because `diff` should be bundled with `cmp` (which is also used in the script) in [diffutils](https://www.gnu.org/software/diffutils/).

Kobzol (approved):
Nice, this also simplifies the script quite a bit. Thanks! I can't really test on OS X, so I'll believe that it helps there :laughing: `tput` is available on Ubuntu, and apparently it should be universally supported.

PR #383, opened 2023-10-28
Store IP address of a submit

### LINKED ISSUE #299: Display IP adress of submit
Display IP address of a machine that uploaded the submit. Also, would be nice to see history of addresses. This is useful for real time tests :-)

DanielKrasny:
Closes #299.
A migration must be run because of the database change.
Since Docker container can only receive client's IP address in host mode, the `X-Forwarded-For` header is used. The IP addresses in the header are delimited by commas, though the real one from nginx [is guaranteed to be the last one](https://nginx.org/en/docs/http/ngx_http_proxy_module.html#var_proxy_add_x_forwarded_for).
Make sure the production nginx proxy has `include proxy_params;` line added after `proxy_pass <docker_image_ip>;`.


Kobzol:
Thanks a lot for working on this! As mentioned in chat, we will postpone this until the end of the semester, it is too dangerous to run migrations during the semester, with the current state of Kelvin :)

templates/web/task_detail.html
```diff
@@ -67,6 +67,7 @@ <h6>
         {{ submit.created_at|timeuntil:submit.assignment.deadline }} after deadline
     </span>
     {% endif %}
+    {% if submit.ip_address %}(from {{ submit.ip_address }}){% endif %}
```
Kobzol, line 70: I would not show this by default here, or at least not to students. Please show it only for teachers.

Kobzol (changes requested):

templates/web/task_detail.html
geordi: I suggest to show only, let's say, first 7 digits of address's SHA256. That should be enough for the comparison.

Kobzol:
Sorry for going back and forth on this :) There are some GDPR problems with storing IP addresses. To summarize what we want to do:
1) Calculate a SHA-256 hash of the IP address of the submit
2) Store only the hash into the DB
3) Display the hash in the task detail (only for teachers)
4) Display the hash in the submit list

DanielKrasny:
Thanks for clarifying. I'll take a look at it later this day.

Kobzol:
Thanks! We will merge this soon, but currently Kelvin has some uWSGI issues :sweat_smile: So let's not add insult to injury by running a DB migration.

geordi:
Thank you all!

PR #387, opened 2023-11-05
Disable default select options

DanielKrasny:
In my opinion, options like `Semester` etc. should not be selected since they are representing some kind of heading. I have set them to `disabled`. We could alternatively use `optgroup` for that, but that's a small detail.

I also found some Czech word in student index template. I have replaced that with this commit because it was just minimal change for a separate commit :)

### DIFF
--- a/frontend/src/ClassDetail.svelte
+++ b/frontend/src/ClassDetail.svelte
@@ -166,7 +166,7 @@ tr td:not(:nth-of-type(1)):not(:nth-of-type(2)):not(:last-child) {
       {#if showStudentsList}
         {#if clazz.summary }
           <button class="p-0 btn btn-link"
-                    on:click={() => showSummary = !showSummary}>{showSummary ? "Hide" : "Show"} exercise summary</button>
+                    on:click={() => showSummary = !showSummary}>{showSummary ? "Hide" : "Show"} class summary</button>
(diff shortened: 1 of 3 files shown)


frontend/src/ClassDetail.svelte
```diff
@@ -166,7 +166,7 @@ tr td:not(:nth-of-type(1)):not(:nth-of-type(2)):not(:last-child) {
       {#if showStudentsList}
         {#if clazz.summary }
           <button class="p-0 btn btn-link"
-                    on:click={() => showSummary = !showSummary}>{showSummary ? "Skrýt" : "Zobrazit"} informace o cvičení</button>
+                    on:click={() => showSummary = !showSummary}>{showSummary ? "Hide" : "Show"} exercise summary</button>
```
Kobzol, line 169:
```suggestion
                    on:click={() => showSummary = !showSummary}>{showSummary ? "Hide" : "Show"} class summary</button>
```
We use "class" for this in Kelvin.

Kobzol (commented):
Apart from the `disabled` attribute and the class naming, it's great. Thanks!

DanielKrasny:
Thank you for the review. I addressed all issues.

Kobzol (approved):
Thanks! :)

PR #388, opened 2023-11-05
Add duplicate buttons to TimeRange

### LINKED ISSUE #386: Make task option setting easier
Setting options for tasks (points, start, deadline) is quite cumbersome for teachers. It would be nice to make it simpler - for example, add buttons to set specific attributes (e.g. deadline) to all lessons. Currently, there is only a button to set all attributes at once to all lessons, and a button to set the points at once. But the same button is not there for start and deadline.

DanielKrasny:
Closes #386.
When a callback is set for duplicate buttons (`onFromDuplicateClick`, `onToDuplicateClick`), these buttons are displayed. The callbacks should accept one argument: the date to be assigned. The buttons can also be disabled using props `fromDuplicateDisabled` and `toDuplicateDisabled`.
The extra props in `TimeRange` are optional and thus they need a default value.

### DIFF
--- a/frontend/src/EditTask.svelte
+++ b/frontend/src/EditTask.svelte
@@ -312,7 +312,7 @@ td:not(:nth-of-type(3)) {
               <td>{ clazz.teacher }</td>
               <td>
                 <div class="row">
-                  <TimeRange timeOffsetInWeek={clazz.week_offset} bind:from={clazz.assigned} bind:to={clazz.deadline} semesterBeginDate={$semester.begin} onFromDuplicateClick={assignAssignedDateToAll} onToDuplicateClick={assignDeadlineToAll} />
+                  <TimeRange timeOffsetInWeek={clazz.week_offset} bind:from={clazz.assigned} bind:to={clazz.deadline} semesterBeginDate={$semester.begin} onFromDuplicateClick={setAssignedDateToVisible} onToDuplicateClick={setDeadlineToAssigned} />
                   <div class="col-md">
                     <div class="input-group">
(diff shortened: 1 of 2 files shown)


Kobzol (commented):
The functionality is awesome, thanks!

I think that we can simplify the code a lot. Let's remove `fromDuplicateDisabled` and `toDuplicateDisabled`, and just enable these buttons inside `TimeRange` if `from` is set to something (if it's not undefined).

Also, let's make `onToDuplicateClick` required and remove the `typeof == 'function'` check, since there's only one usage of `TimeRange` at the moment.

DanielKrasny:
I made both the events (`onFromDuplicateClick` / `onToDuplicateClick`) required. The duplicate buttons in `TimeRange` are always visible now.

frontend/src/EditTask.svelte
```diff
@@ -156,6 +156,22 @@
     }
   }
 
+  function assignAssignedDateToAll(assigned) {
```
Kobzol, line 159:
```suggestion
  function setAssignedDateToVisible(assigned) {
```

frontend/src/EditTask.svelte
```diff
@@ -156,6 +156,22 @@
...
+    task.classes = task.classes.map(cl => {
+      if(isClassVisible(cl))
+        cl.assigned = assigned;
+      return cl;
+    });
+  }
+
+  function assignDeadlineToAll(deadline) {
```
Kobzol, line 167:
```suggestion
  function setDeadlineToAssigned(deadline) {
```

Kobzol (approved):
Thank you! :) Now it works in a reasonable way. I think that the two callbacks could have a slightly better name, but otherwise LGTM.

Kobzol:
I took the liberty of also renaming the usages of those two functions :) Thanks!

Issue #281, opened 2023-02-17
Upgrade or remove dependency on Redis package


geordi:
Since Django 4.0, there's a [Redis backend ](https://docs.djangoproject.com/en/4.1/topics/cache/#redis) built in. Also, current [django-redis-cache](https://github.com/sebleier/django-redis-cache/) we're using for Redis only supports Django up to version 3.2 and latest development stalled in March 2022.

We can use [django-redis](https://github.com/jazzband/django-redis) that is actively developed. It supports newest version of Django and is [still relevant](https://github.com/jazzband/django-redis/issues/585) even that Django supports Redis by itself.

The final option is to simply drop dependency on these packages and use just built-in Redis functionality of Django.

geordi:
Fixed by @Kobzol in #486

PR #391, opened 2023-11-09
Django 5: Packages & Dockerfile

geordi:
This is Draft PR for future upgrade of Django and corresponding libraries in future.

The [django-rq-scheduler](https://pypi.org/project/readwise-django-rq-scheduler/) is unmaintained so we have to use [django-tasks-scheduler](https://github.com/dsoftwareinc/django-tasks-scheduler) in Django 5.

Tasks:
- [x] Migrate to [django-tasks-scheduler](https://github.com/dsoftwareinc/django-tasks-scheduler)
- [x] #281


geordi:
We'll use a different path to upgrade. It's almost done in #486.

PR #396, opened 2023-11-14
EditTask: Use Modal component

DanielKrasny:
I have refactored the use of Bootstrap modal in `EditTask.svelte` to use `Modal.svelte` component, as committed in #375.
After this change, the confirm button in the modal does not have `btn-danger` class but `btn-primary`. If needed, I could add an extra prop to `Modal.svelte` to allow changing button classes.


Kobzol (approved):
Thanks! :)

PR #399, opened 2023-11-17
Add syntax highlighting for Rust

DanielKrasny:


Kobzol (approved):
Oh, I was wondering why the Rust syntax looked so bad :laughing: Thanks! We also have somewhat recently added support for C#. Would you like to send another PR for C# syntax highlighting? :)

PR #403, opened 2023-11-26
Removing students from class

HansS04:
Second attempt at ISSUE #149
I have created a side button with a trash can icon which when clicked will create another column of icons. Click the trash can button on a row with a specific student to remove the link between that student and the course. And that removes the user from the class.

There are also treated states like "is_staff" etc.

Once everything goes well, a message will pop up saying that the user has been successfully removed from the class.

However, there is one problem and that is that I no longer have access to the database and therefore cannot guarantee 100% functionality, so please check that everything is OK. Thanks a lot

### DIFF
--- a/frontend/src/ClassDetail.svelte
+++ b/frontend/src/ClassDetail.svelte
@@ -1,16 +1,17 @@
 <script>
-  import {link} from 'svelte-spa-router'
-  import {fetch} from './api.js'
+  import { link } from "svelte-spa-router";
+  import { fetch } from "./api.js";
 
-  import CopyToClipboard from './CopyToClipboard.svelte'
-  import TimeAgo from './TimeAgo.svelte'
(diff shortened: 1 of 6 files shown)


geordi:
Thanks for your contribution!

frontend/src/ClassDetail.svelte
```diff
@@ -1,275 +1,429 @@
 <script>
-  import {link} from 'svelte-spa-router'
-  import {fetch} from './api.js'
+  import { link } from "svelte-spa-router";
+  import { fetch } from "./api.js";
```
geordi, line 3: I think you used some kind of automatic formatter and it created a lot of changes in this file. Can you just "undo" these changes, so we can easily see what has really changed?

Then just commit it again with: `git commit --amend` and the force push this commit to this branch.

Thank you.

PR #407, opened 2023-12-07
Move test file contents to Svelte

### LINKED ISSUE #397: Foldable inputs in and outputs in tests
Sometimes, inputs and outputs in tests are long, so you have to scroll down or use Page Down key.

Would be nice, if these could be foldable the same way as class is in teacher's view.

DanielKrasny:
Closes #397.
Test files are hidden by default and should be opened manually. The file contents is loaded after the first "unfold" of a card.

### DIFF
--- a/frontend/src/TestFile.svelte
+++ b/frontend/src/TestFile.svelte
@@ -1,0 +1,42 @@
+<script>
+  import SyncLoader from "./SyncLoader.svelte";
+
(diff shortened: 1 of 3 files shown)


Kobzol:
We will solve this once it is rewritten to Vue, it is problematic to resolve in Svelte. Thank you anyway!

PR #420, opened 2024-02-06
Make student view available for teachers

Sawy7:
I propose few changes that aim to make the following possible:

- Teachers, who are also assigned as students in other classes, can now navigate to their _student view_ and therefore access such classes
- When making requests to class related API urls, the server will now return classes where the current user is teacher **or student** (while keeping this endpoint restricted to teachers only)

Let me know whether there are any further changes necessary. Thanks :)


Kobzol:
Thanks! Could you please just squash the commits? :) For example with something like this:
```bash
$ git reset --soft HEAD~2
$ git commit -m"Make student view available for teachers"
$ git push --force
```

geordi:
Thank your for resolving this long standing issue!

PR #422, opened 2024-02-08
Remove link to student-view in student view

geordi:
This removes link to student-view in student view. The link is useless there.

@Sawy7 please, can you look at it and confirm it works for you?


templates/web/layout.html
```diff
@@ -48,18 +48,21 @@
...
-                <li class="nav-item"><a class="nav-link" href="{% url 'admin:index' %}">Admin</a></li>
-                {% endif %}
+                  <li class="nav-item"><a class="nav-link" href="{% url 'submits' %}">Submits</a></li>
+                  <li class="nav-item"><a class="nav-link" href="{% url 'tasks' %}">Tasks</a></li>
+                  {% if request.path != student_index_url %}
+                    <li class="nav-item"><a class="nav-link" href="{% url 'student_index' %}">Student view</a></li>
+                  {% endif %}
+                {%  endif %}
+
+                {% if is_teacher and user.is_staff %}
```
Sawy7, line 61: `is_teacher` state is being checked once before this. I think this could be nested in the previous `if` to make it a bit simpler.

I tested your change locally and it works as expected :+1:

geordi: OK, thanks for spotting this.

PR #430, opened 2024-02-14
Removed usage typing List, Dict, and Tuple types

geordi:
Now uses Python >= 3.9 syntax.


Kobzol (approved):
Not a big fan of this syntax, because some types still need to be upper-cased, so at least before it was consistent, and it also made it clearer what is a type and what is a function/variable name. But since you already changed it, let's go with it.

geordi:
OK, let the consistency prevails.

PR #433, opened 2024-02-16
Fix dark bg color of quick point assignment overlay

Sawy7:
I discovered this hard coded color code for quick point assignment overlay background (mobile feature). With this change it should always pull the correct color for light/dark mode right from Bootstrap, guaranteeing legibility.

Before (dark mode) |  After (dark mode)
:---------------------------:|:---------------------------:
dark-mode-before  |  dark-mode-after


Kobzol (approved):
Huh, I had no idea that Kelvin has such a dialog xD Thanks!

PR #434, opened 2024-02-18
Removing students from class

### LINKED ISSUE #149: Remove of student from class

HansS04:
Closes #149
I have a problem that I no longer have access to the database, so I am asking for a thorough check that everything works and looks as it should. Again, sorry for the delay, I had a lot to do and then I forgot about it ...

### DIFF
--- a/api/urls.py
+++ b/api/urls.py
@@ -19,6 +19,7 @@ urlpatterns = [
     path('transfer_students', views.transfer_students),
     path('semesters', views.semesters),
     path('import/activities', views.import_activities),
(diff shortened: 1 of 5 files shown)


geordi:
@Kobzol I think, we can close it, since it's outdated and doesn't fit to our workflow and we don't want to add new features, right? Sorry @HansS04

Kobzol:
Closing for now, if we will have a use-case for this in the future, we can reopen and rebase.

PR #440, opened 2024-02-28
Add missing import in tests script

DanielKrasny:
A few students reported that there's an error in `run-tests.py` script. I've added missing import of `Optional` to make it working. Tested on Python 3.12.1.


Kobzol (approved):
Weird. Maybe it was removed during the `Optional` -> `|` migration. Thank you! :)

Issue #441, opened 2024-02-28
Assign `teacher` group to newly created users in bulk import


geordi:

geordi:
Solved in #442

PR #442, opened 2024-02-28
Add or update `teachers` group to new teachers (#441)

geordi:


common/bulk_import.py
```diff
@@ -74,6 +74,15 @@ def run(concrete_activities: List[ConcreteActivity], subj: Dict[str, str], semes
                 # TODO: We assign all activities without teacher to one special user :-)
                 teacher = User.objects.get(username='GAU01')
 
+            # At this point, there's only `teachers` group assigned to teachers.
+            # If, the `teacher.groups.get()` has no groups assigned, we know, it's a student and we can assign
+            # `teachers` group to the new user upgrade student to teacher.
+            try:
+                teacher.groups.get()
```
Kobzol, line 81: Why can't we just check if the user has the specific `teachers` group?

geordi: I'll take a look at other way to do it. However, calling `teacher.groups.get()` on a user that is student results in an exception, that's why my solution looks like this.

Kobzol: `teacher.groups.filter(name='teachers').exists()`

Kobzol (approved):

PR #447, opened 2024-03-07
Removed unused imports

geordi:
Also removed or added blank lines to make linter happy.

### DIFF
--- a/api/views.py
+++ b/api/views.py
@@ -1,29 +1,20 @@
 from collections import defaultdict
 
 from django.core import serializers
-from django.shortcuts import render, get_object_or_404
+from django.shortcuts import get_object_or_404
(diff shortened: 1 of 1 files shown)


api/views.py
```diff
@@ -622,7 +616,7 @@ def import_activities(request):
         res['count'] = len(res['users'])
     except (ImportException, UnicodeDecodeError) as e:
         res['error'] = ''.join(traceback.TracebackException.from_exception(e).format())
-    except:
+    except Exception:
```
Kobzol, line 619:
```suggestion
    except BaseException:
```
Otherwise this could in theory change the behavior :)

Kobzol (approved):

PR #454, opened 2024-03-25
Add typehints to `all_sources` and `is_source_valid`

geordi:


Kobzol (approved):
Yeah, it's not exactly easy to understand :)

PR #457, opened 2024-04-02
Update Pyright config

geordi:


.github/workflows/pyright.yml
```diff
@@ -1,14 +1,28 @@
...
+jobs:
+  pyright:
+    name: pyright
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: jakebailey/pyright-action@v2
+        with:
+          version: 1.1.356
+          github_token: ${{ secrets.GITHUB_TOKEN }} # You need this
```
Kobzol, line 12: Where are these options documented? I haven't found it at https://github.com/jakebailey/pyright-action.

geordi: I found it here: https://github.com/jordemort/action-pyright?tab=readme-ov-file#example-usage

But we can experiment with removing `GITHUB_TOKEN`.

Kobzol: Hmm, but you're linking to a different pyright-action, right? The one you link is from `jordemort`, but the GHA workflow uses another one from `jakebailey`.

geordi: Yeah, I combined those two. This PR is basically what we discussed a while ago about trying to make Pyright working in a branch and re-iterating. I'll fix it tomorrow.

Kobzol (commented):
CI config looks OK. The error list... less so xD

geordi:
Yeah, I was experimenting a bit with virtual environment config, but it fails on building dependencies. That's something that can be handled in future.

Anyway, the latest Pyright config update seems to be working. Error list is huge...

PR #458, opened 2024-04-03
Pyright virtual environment config

geordi:
At this pyright check fails on building dependencies. Also, building pandas `1.5.3` in the workflow takes a long time. We can upgrade it to `2.x` and speedup the process. I'll take a look at it.


geordi:
OK, it works now and produces even more typing errors... :)

PR #461, opened 2024-04-09
Set site language

DanielKrasny:
I am frustrated by prompts of my web browser thinking that Kelvin is in Danish language, that's why I added a `lang` attribute to `html` tag so it is clear for browsers.


Kobzol (approved):
Interesting, I never got such prompts. Thanks!

PR #467, opened 2024-04-16
Removed possible infinite loop in INBUS integration

geordi:
Various typing enhancements in INBUS integration, code reformat, and import change.

### DIFF
--- a/common/inbus/inbus.py
+++ b/common/inbus/inbus.py
@@ -57,7 +57,7 @@ def subject_versions(department_id: dto.DepartmentId = dto.DepartmentId(386)) ->
 
                 subject_versions.append(subject_version)
 
-            results: int = len(subject_versions_json)
+            results = len(subject_versions_json)
 
             offset += results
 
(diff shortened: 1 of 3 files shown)


common/inbus/inbus.py
```diff
@@ -21,42 +21,50 @@ def person_by_login(login: str) -> dto.PersonSimple | None:
...
+                subject_version = dto.SubjectVersion(subjectVersionId=subject_version_json['subjectVersionId'],
+                                                     subject=subject,
+                                                     subjectVersionCompleteCode=subject_version_json['subjectVersionCompleteCode'],
+                                                     guarantee=subject_version_guarantee)
 
-            subject_versions.append(subject_version)
+                subject_versions.append(subject_version)
 
-        results = len(subject_versions_json)
+            results: int = len(subject_versions_json)
```
Kobzol, line 60:
```suggestion
            results = len(subject_versions_json)
```
It's a matter of taste, but I find it to be a bit distracting to add typing annotations for obvious types (e.g. return value of `len`) for local variables. In a (reasonable) statically typed language, these annotations would be inferred anyway.

Kobzol (approved):

common/inbus/inbus.py
geordi: Yeah, I got carried away by adding typing annotations :-)

PR #473, opened 2024-09-04
Resolve ruff errors and improve linting & format checking on CI

Kobzol:
Enforce that `ruff` lints and format checks pass on CI.


geordi (approved):
Nice!

PR #476, opened 2024-09-04
Manage packages using the uv package manager

Kobzol:
This PR introduces https://github.com/astral-sh/uv as a package manager. It will allow us to manage packages in a much easier way (similarly to `cargo` with Rust), and also keep a deterministic lockfile, so that we can be sure that all dependencies are exactly as we expect them to be.

I went through the production virtualenv and cross-checked the versions of all packages. They should be the same as in the generated lockfile (`setuptools` is missing in the `uv` lockfile, hopefully that won't be an issue...).


geordi (approved):
We also need to change the deploy script, right?

Kobzol:
Hopefully not :) Let's see.

PR #485, opened 2024-09-05
Update Django to 3.2

Kobzol:
I also updated two packages, which we can only probably test on production (mainly CAS). CAS will have a new migration, which should be automatically applied on production.


geordi (commented):
Also `common/inbus/views.py` uses `HTTPResponse` the same way and needs to be fixed.

Kobzol:
Thanks, fixed.

Kobzol:
Here goes nothing.

PR #490, opened 2024-09-06
Rewrite README of Kelvin and add basic developer documentation

Kobzol:
Also simplify the provided Docker compose file to only provide the most useful things needed for local development.

The rendered Mermaid diagram can be found [here](https://github.com/mrlvsb/kelvin/blob/023a9e992f0ab360277725720c75ec1c4cf0be3d/docs/architecture.md).

### DIFF
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -1,0 +1,90 @@
+# Kelvin development
+Kelvin has uses a relatively complicated set of runtime services, so you need to configure a bunch
+of things to start using it.
+
+## Installing Python dependencies
+First, you need to prepare a Python virtual environment and install dependencies of Kelvin, which is
+a Django app. The easiest way of doing that is to use [uv](https://github.com/astral-sh/uv),
+which manages the dependencies of this project. As a first step for working
+with the Python code, install `uv` using some supported approach, for example:
+
+```bash
+$ pip install --user uv==0.4.4
+```
+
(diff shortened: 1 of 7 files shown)


docs/setup.md
```diff
@@ -0,0 +1,90 @@
...
+of things to start using it.
+
+## Installing Python dependencies
+First, you need to prepare a Python virtual environment and install dependencies of Kelvin, which is
+a Django app. The easiest way of doing that is to use [uv](https://github.com/astral-sh/uv),
+which manages the dependencies of this project. As a first step for working
+with the Python code, install `uv` using some supported approach, for example:
+
+```bash
+$ pip install --user uv==0.4.4
```
geordi, line 12: Current version is `0.4.5`.

geordi (approved):
OK, let's do it.

docs/setup.md
Kobzol: Sure (I think that I started yesterday in the morning with `0.4.4` and in the evening it was `0.4.5` :laughing:), but we use `0.4.4` everywhere currently, and also in production. If we want to update it, we have to do it across the board (ideally in a separate PR).

Kobzol:
Production doesn't use the Docker deployment and it does not have `.env`, so this should hopefully have no effect on it.

PR #491, opened 2024-09-06
Update Psycopg to version 3

Kobzol:
This should provide performance improvements and also unblock future upgrades to new Django versions.


geordi (approved):
LGTM!

PR #495, opened 2024-09-07
Add API for submitting tasks

Kobzol:
I want this functionality to create `cargo kelvin` for making it simpler to upload (Rust) submits into Kelvin.

### DIFF
--- a/common/submit.py
+++ b/common/submit.py
@@ -1,3 +1,4 @@
+import datetime
 import hashlib
 
 from django.contrib.contenttypes.models import ContentType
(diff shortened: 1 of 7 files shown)


common/submit.py
```diff
@@ -11,6 +12,18 @@
 from common.upload import upload_submit_files
 from common.utils import is_teacher
 
+# How much time has to be elapsed between two consecutive commits.
```
geordi, line 15: commits -> submits ?

geordi (approved):
LGTM!

common/submit.py
Kobzol: Thanks! Fixed.

PR #496, opened 2024-09-07
[Fix] SemesterId in INBUs API

geordi:
This has to be elaborated on more thoughly in future. Prefarably syncing Kelvin semester with INBUS's ones in frontend selection. But it also waits for the new frondend implementation.


Kobzol:
Ok, let's hotfix this for now :)

PR #498, opened 2024-09-08
Add the ability to mark a semester as active

Kobzol:
This will help us prepare lessons before a semester starts, without having to offset the beginning weeks.


common/models.py
```diff
@@ -72,6 +85,7 @@ def path_to_code(path):
         return os.path.relpath(path, os.path.abspath("./tasks"))
 
     def dir(self):
+        print(os.getcwd())
```
geordi, line 88: This is probably not needed.

geordi (approved):
LGTM!

common/models.py
Kobzol: Ah, this is a local change that I used to debug the weird getcwd changes, thanks.

PR #500, opened 2024-09-15
Replace `pandoc` binary with `markdown` Python package

Kobzol:
To avoid dependence on a global binary.

I was quite surprised locally that Markdown wasn't rendering :) I was missing the `pandoc` binary.

It seems that it should be mostly compatible with `pandoc` output, although if Markdown was used inside HTML tags, these tags will need to be marked with `markdown="1"` now.


geordi (approved):
LGTM!

Issue #503, opened 2024-09-19
Docker deployment


Kobzol:
We would like to have the option to deploy Kelvin fully inside Docker, ideally with a single command, if possible. We want to have the following services running inside Docker, networked together:

- The Django backend
- nginx, which will serve the Django backend
- Postgres (DB)
- Redis (cache)
- A set of Django RQ workers

This corresponds to the [architecture](https://github.com/mrlvsb/kelvin/blob/master/docs/architecture.md) described in docs.

Ideally, it should be possible to deploy everything with a single `docker-compose.yml` file. All configuration (directory/file paths, ports etc.) should ideally be configurable in the docker-compose file, through environment variables loaded from an `.env` file. You can find an example of that in the existing [docker-compose.yml](https://github.com/mrlvsb/kelvin/blob/master/docker-compose.yml) file.

Here is a broad TODO list of things (in almost arbitrary order) that we need to do in order to make this possible:
- [ ] Add nginx to `docker-compose.yml`
  - [ ] Make it possible to map a host directory that contains nginx config
  - [ ] Make it possible to map a host directory that contains certificates
- [ ] Make it possible to map a host directory containing persistent data for the Redis instance
- [ ] Make sure that all the services in the docker compose file can talk to each other through the network
- [ ] Build the JS frontend in the Kelvin Dockerfile, to make it available in the Docker image
  - [ ] Use a multi-stage build to only include the `frontend.js`, `frontend.css` files and the `dolos` directory in the final Docker image
- [ ] Make it possible to map a host directory containing `local_settings.py`, which will be used to override configuration for the Django backend running inside of Docker
- [ ] Setup nginx so that it serves the Kelvin Django backend
- [ ] Configure a startup script that will run `python3 manage.py migrate` everytime the whole Docker deployment starts
- [ ] Make it possible to start RQ workers inside Docker
  - [ ] Make it possible to run each worker in multiple instances
  - [ ] Workers use Docker internally, configure Docker-in-Docker. An example can be found [here](https://github.com/mrlvsb/kelvin/blob/1f96ae303fc3c61e76c56e1c076bac0c8940393c/docker-compose.yml), but it needs to be tested if it works, and how it interacts with Docker permissions.

If there is a better way to do this, other than docker-compose, we can also try it. But please no Kubernetes :)

Aljallaf100:
I think this will help you.
[...]

PR #504, opened 2024-09-21
Docker Deployment for Kelvin core

JersyJ:
## Issue
Part of #503. 

## Summary
Introduces ability to deploy the Kelvin application (except Django RQ workers) inside Docker with `docker-compose.yml` (profile "prod"). The deployment includes the following networked services:

- The Django backend
- nginx, which will serve the Django backend
- Postgres (DB)
- Redis (cache)

> Added pre-commit
> Introduced additional variables for deployment configuration
> Created Dockerfile for the core app
> Refactored docker-compose.yml for local development and production deployment with profile "prod" (docker compose --profile prod up)
> Created asgi.py for ASGI web servers
> Added CSRF trusted domains for reverse proxy (NGINX) usage
> Included example NGINX configuration

## Testing
Needs Proper Testing! I have tested the basics. Someone with more experience should conduct thorough tests.

## Nginx
To run Nginx, you need to place `privkey.pem` and `fullchain.pem` in the appropriate folder. For testing purposes, you can generate them via `openssl` : 
`openssl req -x509 -nodes -days 365 -newkey rsa:2048 \
  -keyout ./privkey.pem \
  -out ./fullchain.pem \
  -subj "/C=US/ST=YourState/L=YourCity/O=YourOrganization/CN=localhost"`. 

## Remaining Tasks:

- The setup of Django RQ workers is missing


docker-compose-local.yml
```diff
@@ -0,0 +1,34 @@
+# For local development, only database and redis is running
+#
+# docker compose -f docker-compose-local.yml up -d
+#
+
+services:
```
Kobzol: I wanted to avoid having two docker-compose files to avoid unnecessary duplication. I think that we don't need it, because we can start just selected services from the main compose file, right? E.g. locally, the user can just run `docker-compose up [SERVICE1] [SERVICE2]`. Or even better, we could use [profiles](https://docs.docker.com/compose/how-tos/profiles/), and assign a "local" profile only to Postgre + Redis, so that these two can be started easily during development, without the rest.

docker-compose.yml
```diff
@@ -1,18 +1,66 @@
...
-      - POSTGRES_PASSWORD=${DB_PASSWORD?PostgreSQL password}
-  redis:
-    image: redis
+      - POSTGRES_DB=${DATABASE__DB}
+      - POSTGRES_USER=${DATABASE__USERNAME}
+      - POSTGRES_PASSWORD=${DATABASE__PASSWORD}
+    env_file:
+      - .env
+    volumes:
+      - kelvin_db_data:/var/lib/postgresql/data
```
Kobzol: Could you please keep the previous configurable path (`${KELVIN_DB_PATH?Absolute path to Kelvin database storage}:/var/lib/postgresql/data`) so that we can specify the name of the directory where the DB data will be stored? Thanks!

docker-compose.yml
```diff
@@ -1,18 +1,66 @@
...
+      - "${DATABASE__PORT}:5432"
+    
+  redis:
     container_name: kelvin_redis
+    image: redis:alpine
+    restart: unless-stopped
+    env_file:
+      - .env
+    volumes:
+      - "kelvin_redis_data:/data"
```
Kobzol: Same here, so that we can configure the path where Redis stored the data on the host.

docker-compose.yml
```diff
@@ -1,18 +1,66 @@
...
+  nginx:
+    image: nginx:alpine
+    container_name: kelvin_nginx
+    depends_on:
+      - app
+    ports:
+      - "80:80"
+      - "443:443"
+    volumes:
+      - ./nginx.conf:/etc/nginx/nginx.conf:ro
```
Kobzol: And same here, so that we can configure the paths to the config, the certificate directory and the logs directory.

kelvin/settings.py
```diff
@@ -31,9 +31,9 @@
 SECRET_KEY = "***REMOVED***"
 
 # SECURITY WARNING: don't run with debug turned on in production!
-DEBUG = True
+DEBUG = bool(os.environ.get("DEBUG", default=0))
```
Kobzol, line 34: We can keep these two settings as they were, they will be overwritten on production anyway with a `local_settings.py` file (that somehow needs to find its way into the Docker image :) ).

Kobzol (commented):
Thanks, this looks great! Left a few comments inline.

Maybe to clarify a bit, I think that we don't want to commit too many specific things about the deployment into git (e.g. the nginx config). We have some configuration files on the existing server and we'll likely port them over manually, and these will differ on each specific production deployment anyway. The goal is to have `Dockerfile` + `docker-compose.yml` prepared in a way so that it can be deployed easily, and that it can be configured, i.e. I can pass it nginx config/certificates/log dir, `local_settings.py`, paths where Postgre and Redis data should be stored etc.

Kobzol:
Uhh, I only reviewed the first commit, sorry :) I'll take a look at the rest.

kelvin/settings.py
```diff
@@ -149,6 +157,9 @@
 STATIC_URL = "/static/"
 STATIC_ROOT = os.path.join(BASE_DIR, "static")
 
+MEDIA_URL = "/media/"
```
Kobzol, line 160: Where did this come from? :thinking:

JersyJ:
Added pre-commit to avoid future CI issues before pushing to remote.

kelvin/urls.py
```diff
@@ -36,7 +37,7 @@
     path("webpush/", include("webpush.urls")),
     # For django-tasks-scheduler
     path("scheduler/", include("scheduler.urls")),
-]
+] + static(settings.STATIC_URL, document_root=settings.STATIC_ROOT)
```
Kobzol, line 40: Why was this change needed?

kelvin/settings.py
```diff
@@ -149,6 +157,9 @@
...
+MEDIA_ROOT = os.path.join(BASE_DIR, "media")
```
JersyJ, line 161: Should be removed

Dockerfile
```diff
@@ -1,22 +1,75 @@
-FROM python:3.12
+FROM ghcr.io/astral-sh/uv:python3.12-bookworm AS build-backend-stage
```
Kobzol, line 1: The Dockerfile looks very professional :laughing: I wonder if it has to be so complex though.

Could we have just two stages (first would build the frontend, and the second the backend)? It seems a bit overkill to install the `apt` deps twice and use cache mounts, I don't think we need all that here. `uv` is super fast anyway.

docker-compose-local.yml
JersyJ: Okay, no problem! It was just from experience that I’ve noticed people don’t read the documentation and start the Docker Compose without hesitation :D

Dockerfile
```diff
@@ -1,22 +1,75 @@
...
+
+WORKDIR /app
+
+# Create new user to run app process as unprivilaged user
+RUN groupadd --gid 1001 uvicorn && \
+    useradd --uid 1001 --gid 1001 --shell /bin/false --system uvicorn
+
+RUN chown -R uvicorn:uvicorn /app
+
+COPY --from=build-backend-stage --chown=uvicorn:uvicorn /app .
```
Kobzol, line 47: Can't we do just `COPY . .` to avoid all this? E.g. something like this:
```dockerfile
COPY project.toml uv.lock .
RUN uv sync

COPY --from=frontend frontend/frontend.[js|css]|dolos

COPY --chown=uvicorn:uvicorn . .

...
```
In this way if frontend changes, we don't need to rebuild Python deps, but the Dockerfile should be much simpler. I don't think that we need to optimize the fast local Dockerfile rebuild case anyway though.

Kobzol (commented):
Ok now I saw all the commits and it makes more sense :laughing: Sorry.

In general, this already has pretty much almost everything that I wanted, so that is awesome :) It would be nice to de-duplicate the docker-compose files and maybe simplify the Dockerfile a bit (it seems a bit overkill to me at the moment), but otherwise this looks really good.

I hope that the `asgi` changes won't affect the normal `wsgi` operation on the current production deployment, especially since the semester starts in two days :laughing:

docker-compose-local.yml
Kobzol: Well, good point :) But in that case I don't think that people could even conjure the correct `docker-compose -f <XYZ>` command and find the right compose file, so :) And we have like 3 people contributing to Kelvin anyway, so this is not a big concern :D

Dockerfile
JersyJ: This first stage is there for clean runtime, First stage is also installing some dependencies that are not needed when the app is running (libsasl2-dev, libgraphviz-dev), runtime is only graphviz. 

We can remove the cache mount due to old docker build compatibility.

Dockerfile
JersyJ: I did this on purpose to find out which folders are not needed, so you can comment :D, but yes we can copy the whole folder.

docker-compose-local.yml
JersyJ: Done, created profile "prod", local development is still just `docker compose up -d`

kelvin/settings.py
JersyJ: Reverted

kelvin/settings.py
JersyJ: Removed

kelvin/urls.py
JersyJ: Debug purposes, forgot to remove, removed

Dockerfile
JersyJ: Cache mount removed

Dockerfile
```diff
@@ -1,22 +1,67 @@
...
-#RUN addgroup --gid 1000 user
-#RUN adduser --disabled-password --gecos '' --uid $HOST_USER_ID --gid 1000 user
-#USER user
+WORKDIR /frontend
+
+COPY frontend .
+
+RUN mkdir -p /web/static
+
+RUN npm run build
```
Kobzol, line 26: I think that we should have `npm ci` before `npm run build` + add `node_modules` to `.dockerignore`? I think that this now only works because it copies `node_modules` from the host.

nginx.conf
```diff
@@ -0,0 +1,51 @@
+events {
```
Kobzol, line 1: Could you please move this to a `deploy` directory? It probably doesn't hurt if we commit an example nginx config, but let's not put it into the root. And let's also modify the default path in `.env.example` to point to the `deploy` directory.

Dockerfile
```diff
@@ -1,22 +1,67 @@
...
+RUN groupadd --gid 1001 uvicorn && \
+    useradd --uid 1001 --gid 1001 --shell /bin/false --system uvicorn
+
+RUN chown -R uvicorn:uvicorn /app
+
+COPY --from=build-backend --chown=uvicorn:uvicorn /app .
+ENV PATH="/app/.venv/bin:$PATH"
+
+COPY --chown=uvicorn:uvicorn web ./web
+COPY --from=build-frontend --chown=uvicorn:uvicorn /web/static/frontend.[js|css]|dolos ./web/static/
```
Kobzol, line 51: Sorry, this was just pseudocode :laughing: The regex can't be used like this in Dockerfile, it needs to be expanded manually (AFAIK).

Kobzol (commented):
I tried to deploy the website locally using Docker compose and it actually works!! :heart: That's really awesome :laughing:

Left one more issue and two nits.

When I build the image locally, it prints this:
```
useradd warning: uvicorn's uid 1001 is greater than SYS_UID_MAX 999
```
Maybe we should use an UID below 1000, just to be sure?

Dockerfile
Kobzol: Ok, let's keep it, at least we'll make sure that there's no ballast being included in the image.

Dockerfile
JersyJ: True. Done

nginx.conf
JersyJ: Done

Dockerfile
JersyJ: You are right, I did not even check it my bad :D.

JersyJ:
> Maybe we should use an UID below 1000, just to be sure?

True, yes we should. Changed. 

Also changed to WSGI just to be sure, since I looked through the views.

Also moved the ruff.toml to pyproject.toml, to get rid of files in the root of the project.

Kobzol:
Thank you very much for working on this :) This looks really great now. Since it's the first week of the semester, I will delay merging until the weekend, because this also has some changes that could in theory affect the production deployment, which is.. brittle, to say the least :)

Kobzol:
We had some issues with Kelvin over the weekend and had to do a bunch of reverts, so I didn't merge this yet. I'll try it once Kelvin has a bit less traffic.

geordi:
We can also "test deploy" it on the development server. Just saying.

JersyJ:
If you want some help setting up a proper deployment, feel free to reach out to me.

docker-compose.yml
```diff
@@ -1,18 +1,77 @@
...
+      - DATABASE__PASSWORD=${DATABASE__PASSWORD}
+      - REDIS__HOST=redis
+      - REDIS__PORT=6379
+    env_file:
+      - .env
     volumes:
-      - ${KELVIN_DB_PATH?Absolute path to Kelvin database storage}:/var/lib/postgresql/data
+      - app_static:/app/static
+      - app_socket:/socket
+      - ${KELVIN__LOCAL_SETTINGS_PATH}:/app/kelvin/local_settings.py:ro
```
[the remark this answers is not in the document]
JersyJ: That happens only if the file does not exist. Maybe I should have also committed the example of the file. Should I?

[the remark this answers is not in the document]
Kobzol: Right, that makes sense. Ok, let's commit an empty file into the `deploy` directory, and just add some comment like `# This file can be used to override settings for deployment`.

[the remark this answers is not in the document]
JersyJ: Done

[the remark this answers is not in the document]
JersyJ: After forcing the commit with the file, I also did reset of git index to assume-unchanged so it won't be tracked.

uv.lock
```diff
@@ -610,6 +668,22 @@ wheels = [
     { url = "https://files.pythonhosted.org/packages/45/07/95c6be6710207f4a2c83023f59889a86c58282c8367aa49a907e9e6deb57/plum_dispatch-2.2.2-py3-none-any.whl", hash = "sha256:d7ee415bd166ffa90eaa4b24d7c9dc7ca6f8875750586001e7c9baff706223bd", size = 30137 },
 ]
 
+[[package]]
+name = "pre-commit"
+version = "3.8.0"
+source = { registry = "https://pypi.org/simple" }
+dependencies = [
+    { name = "cfgv" },
```
Kobzol, line 676: Why the heck does `pre-commit` have all these dependencies.. sigh.

Kobzol:
I took the liberty of re-creating the `requirements.txt` file and removing the `asgi.py` file, which should not be needed, to reduce the change of causing some issues on production. Let's see if everything still works after this is merged :)

Kobzol (approved):
Thank you very much!

Kobzol:
Looks like everything went through, yay :tada: The next step is for someone to SSH to the new server and try to deploy Kelvin there using Docker.. :)

uv.lock
JersyJ: Pre-commit is powerful tool and is designed to handle much wider range of tools and use-cases. I added it as a dev dependency, so it shouldn’t be an issue.

Kobzol: We currently install deps using `uv pip sync` because of some legacy reasons, so it installs everything :sweat_smile: But it doesn't really matter.

JersyJ:
> I took the liberty of re-creating the `requirements.txt` file and removing the `asgi.py` file, which should not be needed, to reduce the change of causing some issues on production. Let's see if everything still works after this is merged :)

Okk, maybe the next step would be to get rid of the requirements.txt after switching to Docker deployment :)

uv.lock
JersyJ: In my Dockerfile I added parameter which excludes them. So after you switch , no more dev dependencies 😅

Kobzol:
Yeah, of course :) `requirements.txt` is there only as an artifact to keep compatibility with our old virtualenv, which I did not want to recreate. Once we use `uv` full and/or switch to a Dockerized deployment, we should nuke it.

PR #510, opened 2024-09-26
Added support for Vue + Typescript and added eslint + prettier

patrick11514:
Added support for using Vue and Typescript in frontend. 
Added Prettier + Eslint for linting and formatting frontend files.

### DIFF
--- a/frontend/rollup.config.js
+++ b/frontend/rollup.config.js
@@ -1,10 +1,13 @@
-import svelte from 'rollup-plugin-svelte';
-import resolve from '@rollup/plugin-node-resolve';
 import commonjs from '@rollup/plugin-commonjs';
-import livereload from 'rollup-plugin-livereload';
+import resolve from '@rollup/plugin-node-resolve';
+import replace from '@rollup/plugin-replace';
 import terser from '@rollup/plugin-terser';
-import postcss from 'rollup-plugin-postcss'
 import autoprefixer from 'autoprefixer';
+import livereload from 'rollup-plugin-livereload';
+import postcss from 'rollup-plugin-postcss';
+import svelte from 'rollup-plugin-svelte';
+import typescript from 'rollup-plugin-typescript2';
+import vue from 'rollup-plugin-vue';
 import sass from 'sass';
 
 const production = !process.env.ROLLUP_WATCH;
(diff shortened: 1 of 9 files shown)


frontend/rollup.config.js
```diff
@@ -43,6 +46,19 @@ export default {
             emitCss: true,
 		}),
 
+		vue({
+            preprocessStyles: true,
+        }),
+        // Replace the environment variables - https://github.com/rollup/rollup/issues/805
+        // btw the vue/dist/vue.esm.browser didn't work
+        replace({
+            'process.env.NODE_ENV': JSON.stringify('development'),
```
Kobzol, line 55: This should be `"production"` when `production` is true, right? It does seem to save ~30 KiB from the resulting JS archive.

Kobzol (commented):
Thank you! I will probably make a bunch of changes to the TS config in a follow-up PR where I'll configure it on CI, but that does not need to block this PR.

Left one nit, otherwise LGTM.

frontend/rollup.config.js
patrick11514: Yes, it should by production, my bad, also it can be dynamic based on production value.

geordi:
I'm a bit sad that [my branch](https://github.com/mrlvsb/kelvin/tree/vuejs) was kind of ignored, but integration into existing Rollup is nice.

patrick11514:
> I'm a bit sad that [my branch](https://github.com/mrlvsb/kelvin/tree/vuejs) was kind of ignored, but integration into existing Rollup is nice.

Actually I was using your branch, but then we decided with @Kobzol to implement it in current Rollup setup next to Svelte for smoother transition from Svelte to Vue web components.

geordi:
> > I'm a bit sad that [my branch](https://github.com/mrlvsb/kelvin/tree/vuejs) was kind of ignored, but integration into existing Rollup is nice.
> 
> Actually I was using your branch, but then we decided with @Kobzol to implement it in current Rollup setup next to Svelte for smoother transition from Svelte to Vue web components.

OK, no problem at all and thank you for the clarification. It's definitely easier to do transition one technology at time so we don't break things.

I'm VERY happy we can move forward with Vue integration! Thanks!

Kobzol:
The production bundle was only increased by ~30 KiB in production mode, that's quite nice. And today is Friday, so what better time to ship stuff :)

Kobzol (approved):

PR #515, opened 2024-09-27
Swap rollup with vite for building and watching code, added support for Vue, linting

patrick11514:
Current setup with rollup was working, but it has some flaws.
For example:
rollup-plugin-vite was designed to work with rollup-plugin-typescript, which was deprecated and people should use @rollup/plugin-typescript instead, which doesn't work with vue plugin, so people created fork called rollup-plugin-typescript2. Because content of <script setup lang="ts"> was processed by typescript plugin, it can't import vue files, because it doesn't know them, but with some tweaks I successfully fixed this issue, but then another came in. When I was using watch mode, the files (primarly Vue files) which was using typescript plugin was somehow cached and changes doesn't show up in build files, even If i used the @vitejs/plugin-vue, because this issue https://github.com/ezolenko/rollup-plugin-typescript2/issues/433#issuecomment-2075674503.

So we decided to switch to Vite completely.

Hopefully everything will work.  🙏

Changes:
- Swapped Rollup with Vite

## Rollup
832K	frontend.js
288K	frontend.css

## Vite
832K    frontend.js
288K    frontend.css

- because Vite uses Rollup under the hood, bundle sizes seems same

## Vite + Vue (+ example component)
872K    frontend.js
288K    frontend.css

- Added support for Vue
- Added eslint + prettier


geordi:
Thank you! I'll try to test it throughout the weekend.

Kobzol (approved):
Everything looks good. This even updates to Svelte 4, but that should hopefully be also fine.

Attempt number 3.

Kobzol:
Everything looks fine :tada:

geordi:
Thanks @patrick11514

PR #526, opened 2024-10-05
Use Vue component in all tasks page

patrick11514:
Reworked all_tasks page on Teacher side
- Added new created_at column for Task
- Added new endpoint /api/tasks/all, which supports sorting, ordering and filtering tasks
- Created new component, which uses datatable to show, filter and paginate tasks for better clarity


api/views.py
```diff
@@ -56,6 +56,26 @@ def tasks_list(request):
         )
     return JsonResponse({"tasks": result})
 
+def tasks_list_all(request: HttpRequest):
```
Kobzol: This view should be teacher-only, so please mark it with `@user_passes_test(is_teacher)`. It would also be great if we could unify it with `tasks_list` (we just need to make sure that it returns data in a format that is compatible with existing users of `tasks_list`, and make it a bit more REST-friendly, by moving the subject parameter into the URL, i.e. `/tasks/<subject>` (where the subject would be optional).

api/views.py
```diff
@@ -62,19 +62,28 @@ def tasks_list_all(request: HttpRequest):
 
     count = None
     start = None
+    order = ["-created_at", "-id"]
```
Kobzol, line 66: `order = ("-created_at", "-id")`, otherwise it looks like there is a reason why it is mutable, e.g. because you try to append to it somewhere :)

api/views.py
```diff
@@ -62,19 +62,28 @@ def tasks_list_all(request: HttpRequest):
...
 
     if "subject" in request.GET:
         filters["subject__abbr"] = request.GET["subject"]
     if "count" in request.GET:
         count = int(request.GET["count"])
     if "start" in request.GET:
         start = int(request.GET["start"])
+    if "sort" in request.GET:
```
Kobzol, line 74: `if request.GET("sort") == "asc": ...`, we don't need the `else` branch, as it just sets the same default.

frontend/src/utilities.ts
```diff
@@ -36,3 +40,30 @@ export const getFromAPI = async <$ReturnType>(url: string, data?: unknown, metho
...
     return Array.from({ length: size }).map((_, i) => i + start);
 };
+
+/**
+ * Generate string date format from date
+ * @param date Date object or string in format 'YYYY-MM-DD' or 'YYYY-MM-DDTHH:MMZ'
+ * @param withTime Include time in the output
+ * @returns If withTime is false 'YYYY-MM-DD', otherwise 'YYYY-MM-DD HH:MM'
+ */
+export const getDate = (date: Date | string, withTime = false) => {
```
Kobzol, line 50: Let's use a library for this, e.g. https://date-fns.org/. No need to reimplement these kinds of hairy logic by hand, we want to get rid of similar hacks from Kelvin.

web/urls.py
```diff
@@ -133,7 +133,6 @@
         "task/<int:task_id>/csv", teacher_view.download_csv_per_task, name="download_csv_per_task"
     ),
     path("tasks", teacher_view.all_tasks, name="tasks"),
-    path("tasks/<str:subject__abbr>", teacher_view.all_tasks, name="tasks"),
```
Kobzol, line 136: I wonder if someone used this :) Didn't find it in code though.

frontend/src/Teacher/AllTasks.vue
```diff
@@ -0,0 +1,202 @@
...
+/**
+ * Get tasks from API
+ * @param subject Subject abbreviation
+ * @param count Count of records
+ * @param start Offset of records
+ * @param sort Type of sorting
+ * @param search Search query
+ * @returns Tuple of [total count, tasks]
+ */
+const getTasks = async (
```
Kobzol, line 33: Any reason why not just an async function? :)

frontend/src/Teacher/AllTasks.vue
```diff
@@ -0,0 +1,202 @@
...
+    subjects: Subject[];
+  }>('/api/subjects/all');
+
+  if (data) {
+    subjects.value = data.subjects;
+    loaded.value = true;
+  }
+};
+
+getSubjects();
```
Kobzol, line 91: This doesn't need to happen in this PR, but just an idea: we essentially need to load the subjects before displaying this component. We could use `await getSubjects()` + https://vuejs.org/guide/built-ins/suspense to avoid dealing with missing subject data, to make the code a bit easier.

frontend/src/utilities.ts
```diff
@@ -0,0 +1,69 @@
...
+
+        if (!response.ok) {
+            return undefined;
+        }
+
+        const json = await response.json();
+        return json as $ReturnType;
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    } catch (_) {
+        return undefined;
```
Kobzol, line 30: Let's at least log the error message with `console.error`.

Kobzol (commented):
It would be nice to also have the ability to sort by task name. Otherwise looks good, left a few nits.

The commit history is.. too detailed :) As you used one direction and then went into another. So it would be nicer to rebase and get rid of the vuetify stuff.

api/views.py
patrick11514: I don't know, if url `/tasks/subject` will be possible, since there is already path `/tasks/\<path to task\>` which show details about task: https://github.com/mrlvsb/kelvin/blob/master/api/urls.py#L9

Kobzol: Uhh, the paths are a mess :) Then use something else, e.g. `/task-list/...`, or something like that

frontend/src/Teacher/AllTasks.vue
patrick11514: I personally prefer using arrow functions, no special reason here.

web/urls.py
patrick11514: It was used on old page with all tasks, when teacher choosed subject by clicking at link with specific subject abbr, then he was redirected to for example /tasks/UPR

patrick11514: https://github.com/mrlvsb/kelvin/blob/master/templates/web/all_tasks.html#L5

frontend/src/Teacher/AllTasks.vue
patrick11514: So I don't need to show loading at all? This will make getting datatable instance easier.

web/urls.py
Kobzol: Ok, in that case we can remove it now.

frontend/src/Teacher/AllTasks.vue
Kobzol: Ok.

frontend/src/Teacher/AllTasks.vue
Kobzol: Yes, the component would not even start to render until the async data get loaded, which simplifies the code, because it avoid having to deal with `null`/missing values that are loaded asynchronously and that are required for the component to work.

The issue then of course becomes where does the loader component get moved to. With `Suspense`, it gets moved to the parent component. I'm not sure if it is feasible to use Suspense with the web component method that we use right now, though.

frontend/src/Teacher/AllTasks.vue
```diff
@@ -21,6 +21,7 @@ type RawTask = Omit<Task, 'date'> & {
 };
 
 type SortValue = 'asc' | 'desc';
+type OderColumn = 'created_at' | 'name';
```
Kobzol, line 24: `OrderColumn` :)

frontend/src/Teacher/AllTasks.vue
```diff
@@ -188,27 +187,22 @@ const onChangeSubject = () => {
...
-        <a class="btn btn-secondary btn-sm" :href="`/teacher/task/${props.cellData.id}/moss`">
-          MOSS check
-        </a>
-      </template>
-    </DataTable>
-  </template>
+
+  <DataTable ref="dataTable" class="table table-striped" :columns="columns" :options="options">
+    <template #moss="props">
+      <a class="btn btn-secondary btn-sm" :href="`/teacher/task/${props.cellData.id}/moss`">
```
Kobzol, line 201: This should be `/plagcheck` now, instead of `/moss` (this was an pre-existing bug). We should probably figure out a way of sharing routes between the backend and the frontend somehow.. something like https://github.com/tighten/ziggy.

Kobzol:
Thank you very much! This looks great. I took the liberty of writing a small piece of documentation on top of the component, so that we make it easier to understand what it does (it's quite hard to understand what do the existing Svelte components do and where are they used...).

Kobzol (approved):

PR #531, opened 2024-10-09
Import from INBUS rewrite in Vue

geordi:
Initial attempt to rewrite importing of student from INBUS.

Please, can @patrick11514 and @Kobzol look at it when you have some time?


frontend/src/AppInbusImport.vue
```diff
@@ -0,0 +1,327 @@
+<script setup lang="ts">
+import { ref, watch, onMounted } from 'vue';
+
+import { csrfToken } from './api.js';
+
+class WeekActivity {
```
Kobzol, line 6: Since these are only type definitions, it should be `interface`, rather than `class`. We don't need to create any instances of it.

frontend/src/AppInbusImport.vue
```diff
@@ -0,0 +1,327 @@
...
+    },
+    body: JSON.stringify(req)
+  });
+
+  result.value = await res.json();
+  busy.value = false;
+}
+
+onMounted(() => {
+  watch(subject_inbus_selected, (newValue, oldValue) => {
```
Kobzol, line 228: I'd put the `watch`es onto the top level, to make them more visible. They don't need to be created inside `onMounted`.

frontend/src/AppInbusImport.vue
```diff
@@ -0,0 +1,327 @@
...
+  /**
+    Výskyt aktivity v konkrétním týdnu výuky.
+    */
+
+  weekActivityId: number; // ID výskytu aktivity v konkrétním týdnu výuky
+  weekNumber: number; // číslo týdne v semestru
+  date: string; // datum výskytu aktivity (yyyy-MM-dd)
+}
+
+class ConcreteActivity {
```
Kobzol, line 16: Could you please put these type definitions into a separate file, e.g. `inbus.ts`?

frontend/src/AppInbusImport.vue
geordi: One moved to the top level, one removed.

frontend/src/AppInbusImport.vue
geordi: Done!

frontend/src/Teacher/InbusImport.vue
```diff
@@ -0,0 +1,279 @@
...
+
+let subject_inbus_selected = ref<InbusSubjectVersion | null>(null);
+let subject_kelvin_selected = ref(null);
+
+let semester = ref(null);
+let busy = ref<boolean>(false);
+
+let semesters = ref(null);
+
+let semesters_loading = ref<boolean>(true);
```
Kobzol, line 46: Semesters and subjects are always loaded just once. Let's simplify this by just using a single `loading` variable for them.

frontend/src/Teacher/InbusImport.vue
```diff
@@ -0,0 +1,279 @@
...
+let inbus_and_kelvin_subjects_loading = ref<boolean>(true);
+let inbus_schedule_loaded = ref<boolean>(false);
+
+let subjects_inbus: InbusSubjectVersion[] | null = null;
```
Kobzol, line 50: Local variables should just be inside functions, no need to hoist them to the global scope. `subjects_inbus` should be a local variable.

Also, `subjects_inbus_filtered` and `subjects_kelvin` should be `ref`s, because they are used in the template.

frontend/src/Teacher/InbusImport.vue
```diff
@@ -0,0 +1,279 @@
...
+  created: boolean;
+}
+
+interface Result {
+  users: ImportResult[];
+  count: number;
+  error?: string;
+}
+
+let subject_inbus_selected = ref<InbusSubjectVersion | null>(null);
```
Kobzol, line 38: `subject_inbus_selected` is not needed, you can just read the selected value from a parameter of `loadScheduleForSubjectVersionId`. There's not really any point in combining both `@change` and `v-model`.

frontend/src/Teacher/InbusImport.vue
```diff
@@ -0,0 +1,279 @@
...
+let subject_kelvin_selected = ref(null);
+
+let semester = ref(null);
+let busy = ref<boolean>(false);
+
+let semesters = ref(null);
+
+let semesters_loading = ref<boolean>(true);
+let inbus_and_kelvin_subjects_loading = ref<boolean>(true);
+let inbus_schedule_loaded = ref<boolean>(false);
```
Kobzol, line 48: You can just check if `subject_inbus_schedule` is `null` (and reset it to `null` at the beginning of `loadScheduleForSubjectVersionId`, no need to create another variable for it.

frontend/src/Teacher/InbusImport.vue
geordi: After our discussion. I removed all `loading` variables.

frontend/src/Teacher/InbusImport.vue
geordi: Done.

frontend/src/Teacher/InbusImport.vue
geordi: OK. Now I have to use index into an array that I don't like but we can iterate on that later on.

frontend/src/Teacher/InbusImport.vue
```diff
@@ -0,0 +1,266 @@
...
+  error?: string;
+}
+
+let subject_kelvin_selected = ref(null);
+
+let semester = ref(null);
+let busy = ref<boolean>(false);
+
+const semesters = ref(await loadSemesters());
+const [subjects_inbus_filtered, subjects_kelvin]: [
```
geordi, line 44: I need a hint to annotate it properly or how to rewrite the whole line to make ESLint happy.

Kobzol:
When you use the top-level `await`, the variables no longer need to be `ref`, sorry (I didn't know you're going to switch to top-level await). Only things that changes need to be `ref`s (or `computed` or something like that). If it's read-only, it can be a normal variable.

I refactored the loading code a bit to make it simpler.

geordi (commented):
I tried almost exactly the same way and it didn't work. Maybe I had a bug somewhere else and I scratched the idea. Anyway, now it's much better.

frontend/src/Teacher/InbusImport.vue
```diff
@@ -237,7 +247,18 @@ function onInbusSubjectSelected(event: Event) {
...
 
         <td>
-          {{ ca.teacherFullNames }}
+          <span v-if="ca.teacherFullNames">{{ ca.teacherFullNames }}</span>
+          <span v-else>
+            <select @change="onTeacherSelected">
+              <option
+                v-for="teacher in teachers"
+                :key="teacher.username"
+                :value="[ca.concreteActivityId, teacher.username]"
```
Kobzol, line 256: We should just format it, rather than depending on the serialization of an array. Something like this:
```typescript
:value="`${ca.concreteActivityId},${teacher.username}`"
```

geordi: Thanks, will force push a fix.

frontend/src/Teacher/InbusImport.vue
```diff
@@ -0,0 +1,322 @@
...
+  return teachers;
+}
+
+async function importActivities() {
+  busy.value = true;
+  const req = {
+    semester_id: semester.value,
+    subject_abbr: subject_kelvin_selected.value,
+    activities: classes_to_import.value,
+    activities_to_teacher: activities_to_teacher_selected.value
```
Kobzol, line 184: Note that `activities_to_teacher_selected` is empty by default, it will **NOT** be pre-filled with the first teacher which is then shown in the UI. That means that unless you explicitly select a teacher for the lesson, it will be missing and the bulk import will fail.

Therefore, the backend should either use fallible indexing for it, and still set some default teacher, or the UI should not allow importing the subject until you explicitly select a teacher (or alternatively, the UI should just pre-fill the first teacher from the list automatically).

geordi: OK, hope it's fixed in the current version.

geordi (commented):
I know that time complexity of the current way of dealing with user errors is suboptimal but I hope it's good enough for us.

Kobzol (approved):
Let's ship it :) Good job.

geordi:
Learned a lot in the past few weeks thanks to you @Kobzol

PR #533, opened 2024-10-09
Remove plagiarism SVG graph rendering

Kobzol:
I want to see if this is causing the slow load of some student's submit pages.


Kobzol:
Yup, that was it, now it loads ~instantly.

PR #534, opened 2024-10-09
Remove plagiarism graph from result page

Kobzol:
The SVG generation can be quite slow, this should make both submit and plagiarism page loading much faster.


geordi (approved):
LGTM!

PR #537, opened 2024-10-12
Send an e-mail to student after they get assigned points

Kobzol:


geordi (approved):
LGTM!

Kobzol:
Right now it will just send the e-mails to the console. We can merge it, but maybe it's better to wait until we get the proper SMTP credentials on the production server.

PR #556, opened 2024-11-12
Notifications rewrite + small fixes

### LINKED ISSUE #455: Notifications are not scrollable
When student has too many notifications, to scroll in them, they have to scroll the entire site site. The notification box should be scrollable itself.

image
image

patrick11514:
- Added ability for SuspensionWrapper to pass rest of props to child
- Moved utilities.ts file to folder called utilities and renamed it to utils.ts to maintain some folder structure, that having everything in src folder
- Added getDataWithCSRF, which just fills the header `X-CSRFToken` and then call getFromAPI. This is alternative to fetch function in api.js (also swapped order of parameters of getFromAPI from url, data, method, headers to url, method, data, headers, beacause we usually want to specify method first and then data (if any).
- Ported Notifications and refactored code a little bit + fixed #455 by setting max height to 50vh, if exceeded, notifications list is scollable
- Added ability to switch between old layout (currenly only Notificaions), with cookie newUI. If set (technically to any value), then new Vue components are used, when 0 or not presented, then old Svelte components are used. When notification will be tested in production, then we can just remove the svelte version and remove the condition in front of Vue version.


frontend/src/components/Loader.vue
```diff
@@ -1,63 +1,68 @@
 <script setup lang="ts">
-import { generateRange } from '../utilities';
+import { generateRange } from '../utilities/utils';
```
Kobzol, line 2: When we already have a subfolder named `utilities`, the files/modules inside of it should ideally have clear reponsibilities. I think that here `utils` could be renamed e.g. to `api` or `requests`, and the storage stuff could be moved to a different file called e.g. `storage`. `generateRange` can just be inside `Loader.vue`.

frontend/src/components/Notifications.vue
```diff
@@ -1,4 +1,12 @@
 <script setup lang="ts">
+/**
+ * This components displays bell with number of notifications next to it.
```
Kobzol, line 3: `components` -> `component`

frontend/src/components/Notifications.vue
```diff
@@ -1,4 +1,12 @@
...
+ * After opening, it shows list of all notifications, which user got.
+ * It is available to both teachers and students, and it is accesible from
```
Kobzol, line 5: These two sentences don't make sense together, probably a copy-paste error? :)

Kobzol (commented):
Looks great! Left a few nits, and style/lints need to be fixed.

I realized that there's a lot of duplicated code with Svelte now. Since the component seems to work well, let's just nuke the old Svelte code and notifications code, to avoid having two copies there.

But please still keep the `newUI` logic, we will use it in the future for more complicated components.

Kobzol:
I will do some push notification fixed on top of this PR (https://github.com/mrlvsb/kelvin/issues/557) once this is merged.

patrick11514:
> I realized that there's a lot of duplicated code with Svelte now. Since the component seems to work well, let's just nuke the old Svelte code and notifications code, to avoid having two copies there.

I just looked into svelte files, and we can only remove Notifications.svelte file, becuase notifications.js is also used [here](https://github.com/mrlvsb/kelvin/blob/master/frontend/src/TaskDetail.svelte#L9)

Kobzol:
Just out of curiosity, is Svelte (+ our build system) unable to import code from `.ts` files? Or perhaps use the `.js` prefix when referring to TS files?

Kobzol (approved):
Ok, awesome :) Thank you, this is great!

PR #561, opened 2024-11-28
ColorTheme Rewrite

patrick11514:
(splited #560 into separated Pull Request (Editor + ColorTheme)  as suggested by @Kobzol )

This is rewrite of ColorTheme, which is theme picker in the navigation, next to notification bell. Here you can select if you want auto, light or dark theme.
It was rewriten, because #560 depends on ColorTheme, for changing their theme.

I would personally don't replace old ColorTheme, because current Editor.svelte depends on the old ColorTheme.svelte, so with the new ColorTheme the svelte editor will be white. (can be seen in video attached to #560 )


frontend/src/components/ColorTheme.vue
```diff
@@ -0,0 +1,100 @@
...
+    immediate: true
+  }
+);
+
+if (curTheme.value != 'auto') {
+  selectTheme(curTheme.value);
+}
+</script>
+
+<template>
```
Kobzol, line 74: This component should ideally receive a prop with the theme, and emit an event when it is changed. The rest of the logic should be in a parent component.

Kobzol (approved):
Looks good, thank you! :)

PR #570, opened 2025-01-07
Update list of needed system packages

geordi:


geordi:
I'll test it in Python 3.13 and update Dockerfile if needed.

PR #571, opened 2025-01-07
Upgrade packeges for Python 3.13 runtime

geordi:
To consider while migrating to new VM.


geordi:
We stay at Python 3.12 for the moment.

Issue #575, opened 2025-01-22
Configure renovatebot


Kobzol:
It would be nice to configure [Renovatebot](https://docs.renovatebot.com/) to periodically update Python and Javascript packages in this repository.

PR #576, opened 2025-01-22
Remove dependabot

Kobzol:
It doesn't support uv, and in general lacks in functionality when compared to renovatebot.

https://github.com/mrlvsb/kelvin/issues/575


geordi:
OK

PR #579, opened 2025-01-23
Deploy Docker image to ghcr.io

Kobzol:
This changes the deployment from a webhook to a docker push to the ghcr.io registry. It also switches to merge queues.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,112 @@
+name: CI
+on:
+  pull_request:
+  merge_group:
```
JersyJ, line 4: we don`t use merge queue, use classic push
````
push:
      branches:
        - 'main'
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,112 @@
+name: CI
+on:
+  pull_request:
```
JersyJ, line 3:
```suggestion
  pull_request:
      types: [ opened, synchronize, reopened ]
      branches:
        - 'main'
```

to avoid pull requests triggers into different branches (it would be overkill)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,112 @@
...
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+      - name: Install dependencies
+        run: |
+          sudo apt-get install libgraphviz-dev
+          python -m pip install --upgrade pip
+          pip install uv==0.4.4
+          # Use the legacy lockfile to battle test it
+          uv pip sync --system requirements.txt
```
JersyJ, line 21: install of requirements should be separated and should be in the venv, so the developer has same instructions as the workflow, this reproduces developer workflow and detect possible problems

pros: the cache could be used to speed up this workflow

.github/workflows/ci.yml
```diff
@@ -0,0 +1,112 @@
...
+        with:
+          name: kelvin
+          path: ${{ runner.temp }}/kelvin.tar
+  deploy:
+    runs-on: ubuntu-latest
+    permissions:
+      packages: write
+    needs: [test-backend, test-frontend, build-docker]
+    environment: production
+    if: ${{ github.event_name == 'merge_group' }}
```
JersyJ, line 89:
```suggestion
     if: |
         github.event_name != 'pull_request' &&
         ( github.ref == 'refs/heads/main' )
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,112 @@
...
+        with:
+          cache-from: type=registry,ref=ghcr.io/mrlvsb/kelvin-ci-cache
+          cache-to: type=registry,ref=ghcr.io/mrlvsb/kelvin-ci-cache,compression=zstd
+          tags: ghcr.io/mrlvsb/kelvin:latest
+          outputs: type=docker,dest=${{ runner.temp }}/kelvin.tar
+      - name: Share built image
+        uses: actions/upload-artifact@v4
+        with:
+          name: kelvin
+          path: ${{ runner.temp }}/kelvin.tar
```
JersyJ, line 82: it this really needed? who will download it and change tags in the docker compose and test it?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,112 @@
...
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Download built image
+        uses: actions/download-artifact@v4
+        with:
+          name: kelvin
+          path: ${{ runner.temp }}
```
JersyJ, line 103: there should be another run of build when the PR is merged into main branch

.github/workflows/ci.yml
```diff
@@ -0,0 +1,112 @@
...
+      - name: Load image
+        run: |
+          docker load --input ${{ runner.temp }}/kelvin.tar
+          docker image ls -a
```
JersyJ, line 107: overcomplicated

JersyJ (changes requested):
I believe it does not make sense to run CI on the main branch if you are not evaluating test coverage for the branch or other specific metrics. Running linters is also unnecessary when they are already executed on every commit of a pull request.

In my opinion, deployments should be handled separately and triggered either by a push to the main branch or manually using workflow_dispatch

Dockerfile
```diff
@@ -1,14 +1,7 @@
 FROM ghcr.io/astral-sh/uv:python3.12-bookworm AS build-backend
 
-# Workaround for a timeouting VSB DNS. It seems that it is sadly unable to download APT packages
-# from the default Debian mirror.
-RUN sed -i "s#deb.debian.org/debian\$#ftp.debian.cz/debian#g" /etc/apt/sources.list.d/debian.sources
```
JersyJ, line 5: Is the VSB DNS resolved?

geordi: By now, DNS works.

.github/workflows/ci.yml
JersyJ: sorry but the point stay the same
```suggestion
  pull_request:
      types: [ opened, synchronize, reopened ]
      branches:
        - 'master'
```

Dockerfile
Kobzol: Seems like it might have been a transient issue, but in any case we will build the Docker images on GitHub Actions, and just pull them on the Kelvin server(s), so the workaround shouldn't be needed anymore.

.github/workflows/ci.yml
Kobzol: Yeah, I want to get rid of `requirements.txt`, we don't need it anymore, because the deployed server is now managed by `uv`. But I'll do that in a separate PR.

.github/workflows/ci.yml
Kobzol: It is needed to pass the built Docker image to the `deploy` job. The deployment itself cannot be a separate step in the Docker build job, because it must only be performed if tests (the other two jobs) have succeeded. I also need it to be a separate step to make environments work.

.github/workflows/ci.yml
Kobzol: I prefer not to run any CI workflows after a push to main, if possible. Everything should be ideally done ahead of the merge.

Kobzol:
I added a conclusion job, which is not super pretty, but often a requirement if one wants to get GitHub Actions to work well with both pull request and merge queue workflows.. Anyway, looks like it behaves like I wanted it to now.

PR #604, opened 2025-02-07
Snapshot testing

patrick11514:
This pull request introduces snapshot testing via Vitest of Vue + older Svelte components.

Currently I provided tests for these Vue components:
- Notifications
- ThemeSelector

and these Svelte components:
- Markdown

I will add more test later.

Also this pull request refactored Notifications, because they loaded data on import, which was not best.

Todo:
- [x] Add running tests to CI
- [ ] Add test to AllTasks + TimeAgo


frontend/src/components/Notifications.vue
```diff
@@ -6,30 +6,54 @@
...
+  markAllRead,
+  getPushStatus,
+  type Notification,
+  subscribePushNotifications
 } from '../utilities/notifications';
 import TimeAgo from './TimeAgo.vue';
 
-async function enablePushNotifications() {
-  if (!(await pushNotifications.subscribePushNotifications())) {
+type DePromise<$Promise> = $Promise extends Promise<infer $Inner> ? $Inner : $Promise;
```
Kobzol, line 20: What is this? :)

frontend/src/utilities/notifications.ts
```diff
@@ -24,129 +22,121 @@ export type Notification = {
...
-            .content;
-    }
+const getPublicKey = () => {
+    return document.querySelector<HTMLMetaElement>('meta[name="django-webpush-vapid-key"]').content;
+};
 
-    function urlB64ToUint8Array(base64String: string) {
-        const padding = '='.repeat((4 - (base64String.length % 4)) % 4);
-        const base64 = (base64String + padding).replace(/-/g, '+').replace(/_/g, '/');
+const urlB64ToUint8Array = (base64String: string) => {
```
Kobzol, line 50: :(

frontend/src/utilities/notifications.ts
```diff
@@ -24,129 +22,121 @@ export type Notification = {
...
+            };
+        }
     }
 
     return {
-        ref: pushNotificationsStatus,
-        subscribePushNotifications
+        supported: false,
+        enabled: null,
+        reg: reg as null //because if ServiceWorker or PushManager is not supported, reg will be null
```
Kobzol, line 140: `reg: null`

frontend/vite.config.mjs
```diff
@@ -38,6 +38,11 @@ export default defineConfig((env) => {
             postcss: {
                 plugins: [autoprefixer()]
             }
+        },
+        test: {
+            globals: true,
+            environment: 'jsdom',
+            setupFiles: 'src/utilities/setupTests.ts'
```
Kobzol, line 45: Where is this file?

frontend/src/components/ThemeSelector.test.ts
```diff
@@ -0,0 +1,35 @@
...
+        });
+        expect(wrapper.html()).toMatchSnapshot();
+    });
+    it('Theme prop to light', () => {
+        const wrapper = mount(ThemeSelector, {
+            props: {
+                theme: 'light' satisfies ThemeValue
+            }
+        });
+        expect(wrapper.html()).toMatchSnapshot();
```
Kobzol, line 25: It's not very intuitive from this to understand what the test is supposed to test. I'd expect that we assert that a given item in the select box is selected.

Kobzol (commented):
Left some comments. It's a good start for our frontend tests, thank you!

We should also actually run the tests on CI.

frontend/vite.config.mjs
patrick11514: Oh, I forgot to remove It. I was trying something with mocking theme, so components, which depends on color theme can be properly tested and for example the color theme can be changed, but didn't make it work, so I didn't commited the file.

frontend/src/utilities/notifications.ts
patrick11514: Sorry :( but as I said, I prefer arrow functions, I know that they don't have their own this, and cannot be used everywhere, or with knowledge, that they inherit its this from surrounding scope, but I wanted to unify it with other function in file.

frontend/src/components/Notifications.vue
Kobzol: What about just creating a proper (named) type for the return value of `getPushStatus` and using that in both places? This is unnecessarily complex and hard to understand.

patrick11514: Oh right, thanks, I sometimes over-complicate things 😅

Kobzol:
Thanks! I don't think that we need to add more tests in this PR, but it would be nice to run the tests in CI. Could you please take a look at that?

frontend/src/components/ThemeSelector.test.ts
patrick11514: Something like this?
```TS
it('Theme prop to auto', () => {
        const wrapper = mount(ThemeSelector, {
            props: {
                theme: 'auto' satisfies ThemeValue
            }
        });

        expect(wrapper.get('ul > li:nth-child(1) > button').classes()).toContain('active');
    });
```

Kobzol: Yeah, looks good :) Sadly it's not semantically a `<select>`, where it would be easier to check what is selected.

patrick11514:
Okay I found one issue. https://github.com/mrlvsb/kelvin/actions/runs/13289288861/job/37105646568?pr=604
The new Date in TimeAgo.vue reports different time in CI environment, probably because of timezone? or maybe different time set on runner. So the tests fail :/

patrick11514:
Okay, we can set the timezone to CET instead of default UTC, for the test command using:
```YAML
        env:
          TZ: "CET"
```

.github/workflows/ci.yml
```diff
@@ -44,6 +44,12 @@ jobs:
         run: |
           cd frontend
           npm run check
+      - name: Run tests
```
Kobzol, line 47: Could you please add run tests before running lints? Usually you want to get the most valuable signal first, to avoid a situation where CI fails on formatting/lints before it tells you that tests are actually broken.

Kobzol (approved):
You're priotizing developers from central Europe, but I guess that's fine given our contributors pool :laughing:

Left one comment regarding CI, otherwise LGTM. The notification changes should be split into a separate PR, but let's not deal with that.

Kobzol (approved):
Thanks!

PR #609, opened 2025-02-13
Use INBUS semester to get subject schedule

geordi:
This removes harcoded INBUS semester id.

### DIFF
--- a/common/inbus/inbus.py
+++ b/common/inbus/inbus.py
@@ -88,14 +88,14 @@ def subject_versions(
 
 
 def schedule_subject_by_version_id(
-    subject_version_id: dto.SubjectVersionId,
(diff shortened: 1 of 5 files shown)


common/inbus/inbus.py
```diff
@@ -88,14 +88,14 @@ def subject_versions(
 
 
 def schedule_subject_by_version_id(
-    subject_version_id: dto.SubjectVersionId,
+    subject_version_id: dto.SubjectVersionId, inbus_subject_id: int
```
Kobzol, line 91:
```suggestion
    subject_version_id: dto.SubjectVersionId, inbus_semester_id: int
```

geordi: Thanks for spotting! It needs to be changed also a few lines below as well.

Issue #591, opened 2025-02-03
Java pipeline add java version 21 and 23


david-jezek:
Upgrade docker image configuration for java pipeline to allow Java 21, Java 23

PR #612, opened 2025-02-17
feat(pipeline):add java 21and 23 to dockerfile (#591)

david-jezek:
change only in dockerfile for java runner


Kobzol:
Do you need all of them? We usually just update the version, i.e. change 17 to 23.

david-jezek:
> Do you need all of them? We usually just update the version, i.e. change 17 to 23.

Nope. Created new force-push only with  java 21

Issue #597, opened 2025-02-03
Extends image for pipeline run with JDK


david-jezek:
Add Java 21 (or 23) into pipeline run image to allow run scripts not only in Python but also i Java

PR #613, opened 2025-02-17
feat(pipeline):add java 21 into dockerfile (#597)

david-jezek:
Just add java package to dockerfile for "run" pipeline

### DIFF
--- a/evaluator/images/run/Dockerfile
+++ b/evaluator/images/run/Dockerfile
@@ -1,7 +1,8 @@
 FROM kelvin/base
 
 RUN apt-get update && \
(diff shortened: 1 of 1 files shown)


Kobzol:
Out of curiosity, how does running a Java "script" look like? Do you have a `.jar` that you want to execute with `java -jar`?

david-jezek:
> Out of curiosity, how does running a Java "script" look like? Do you have a `.jar` that you want to execute with `java -jar`?

It is new java feature named "Unamed classe" (https://www.baeldung.com/java-21-unnamed-class-instance-main).
It ca be run java --source 21 --enable-preview HelloWorld.java

Kobzol:
Cool :) That should be indeed useful for Kelvin. Thanks!

PR #622, opened 2025-02-22
Generalization of markdown processing

erikhaw:
Generalization of markdown processing so that it can be used for another group of assets.

### DIFF
--- a/web/markdown_utils.py
+++ b/web/markdown_utils.py
@@ -14,7 +14,7 @@ from kelvin.settings import BASE_DIR
 
 
 class ProcessedMarkdown:
-    def __init__(self, name: str, content: str , announce: str, meta: dict = None):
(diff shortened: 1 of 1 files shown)


web/markdown_utils.py
```diff
@@ -13,11 +13,11 @@
 from kelvin.settings import BASE_DIR
 
 
-class Readme:
-    def __init__(self, name, announce, content, meta=None):
+class ProcessedMarkdown:
+    def __init__(self, name, content, announce, meta=None):
```
Kobzol, line 17: Since we're modifying the code, could you please add type hints to the parameters, and a comment that explains what `announce` does? To leave this in a better place than we found it :) Thanks!

erikhaw: Sure, please check follow up commit.

Kobzol (approved):
Thank you!

Kobzol (approved):

PR #624, opened 2025-02-24
Fix URLs of Clang-Tidy checks documentation

DanielKrasny:
For quite a long time, Kelvin displays Clang-Tidy hints that point to invalid location in the docs because LLVM has changed the directory structure of the docs. I have fixed the script to cover these changes, so the hints can point to relevant docs link again.

Since `clang-tidy` evaluator image depends on statically copied `urls.json`, an image rebuild is required.


evaluator/images/clang-tidy/build_urls.py
```diff
@@ -29,19 +32,64 @@ def get(url):
 def process(check):
```
geordi, line 32: We're trying to put type hints wherever we touch old codebase. Could you do it for `check` parameter, please?

evaluator/images/clang-tidy/build_urls.py
```diff
@@ -16,9 +19,9 @@ def get(url):
     except FileNotFoundError:
         logging.info("Downloading %s", url)
         try:
-            with urllib.request.urlopen(url) as f:
+            with urlopen(url) as f:
```
geordi, line 22: Somehow, I like the longer version with full package path a bit more. What do you think @Kobzol ?

geordi (approved):
LGTM

Kobzol (commented):
Hi, thanks for taking a look at this! I think that we could simplify this whole logic a lot, because the URLs seem to be predictable.

`llvm-include-order` => `https://clang.llvm.org/extra/clang-tidy/checks/llvm/include-order.html`
`abseil-duration-division` => `https://clang.llvm.org/extra/clang-tidy/checks/abseil/duration-division.html`

We can just take the part before the first dash (`-`), put a slash (`/`) after rit, and then use the rest of the name of the check.

That way it should be much simpler to get the proper URL.

Also, to avoid this getting stale again, it would be better if we executed the `get_urls.py` script in `Dockerfile`, so that it is always rebuilt as part of `docker build`.

Let me know if you're interested in working on this.

evaluator/images/clang-tidy/Dockerfile
```diff
@@ -1,4 +1,5 @@
 FROM alpine:edge
 RUN apk update && apk add clang-extra-tools libc-dev linux-headers python3 py3-yaml libstdc++ g++
-ADD analyze.py urls.json /
+ADD analyze.py build_urls.py /
+RUN /build_urls.py && rm -rf /tmp/*
```
Kobzol, line 4: Actually now that we do it in the Dockerfile, the cache is kind of useless :sweat_smile: We can simplify it further and just download the URL everytime in `get`.

Actually, do we even need to download the page? Can't we just used the original URL directly? I don't know why does the page need to be scanned :man_shrugging:

DanielKrasny: You're right with the cache. Let's remove it.

I still believe, though, that the page download has its sense since it resolves aliases and detects invalid pages which are listed as `null` in the `urls.json`. I think it is better that you don't need to be redirected immediately after click.

Kobzol: I would gladly trade a redirect for removing 20 lines of code from Kelvin :laughing: But yeah, let's keep it, you simplified it quite a lot already.

Kobzol (approved):
image

Now this is the kind of PR diffs that I'd like to see :) Thank you very much!

PR #625, opened 2025-02-25
Initialization of quiz app and its models

erikhaw:
Added quiz app and its models with initial migration.


Kobzol:
`COPY --chown=webserver quit ./quit` will need to be added to the Dockerfile. I'll take a look at the rest later.

quiz/models.py
```diff
@@ -0,0 +1,75 @@
...
+        verbose_name_plural = "Quizzes"
+
+
+# Model that represents an assigned quiz
+class AssignedQuiz(models.Model):
+    quiz = models.ForeignKey(Quiz, on_delete=models.CASCADE)
+    clazz = models.ForeignKey(Class, on_delete=models.CASCADE)
+    assigned = models.DateTimeField()
+    # duration of quiz in minutes
+    duration = models.IntegerField()
```
Kobzol, line 29: I can imagine long running quizzes that don't have a duration, but we can always make it nullable in the future, so let's keep it required for now.

quiz/models.py
```diff
@@ -0,0 +1,75 @@
...
+        verbose_name_plural = "Quiz templates"
+
+
+# Model that represents an enrolled quiz
+class EnrolledQuiz(models.Model):
+    assigned_quiz = models.ForeignKey(AssignedQuiz, on_delete=models.CASCADE)
+    template = models.ForeignKey(TemplateQuiz, on_delete=models.CASCADE)
+    student = models.ForeignKey(User, on_delete=models.CASCADE)
+    deadline = models.DateTimeField()
+    max_points = models.FloatField(default=0.0)
```
Kobzol, line 61: It confuses me a bit why `max_points` is here and not in `AssignedQuiz`. Could you elaboreate please? :)

Kobzol (commented):
Thanks! The models look fine, left a couple of questions.

I'm not sure what consequences it wiill have to store the models inside the `quiz` app, because in the future I can imagine that some models (or code) from `common` might want to touch them, which would create a cyclic dependency between `common` and `quiz`. But there are ways in Python to break these dependencies, so let's keep it like this for now.

quiz/models.py
```diff
@@ -0,0 +1,75 @@
+from django.db import models
+from common.models import Class, Subject, User, Semester
+
+
+# Model that represents a quiz
+class Quiz(models.Model):
+    name = models.CharField(max_length=100)
+    # root directory of stored quiz
+    root = models.CharField(max_length=255, default="quizzes")
```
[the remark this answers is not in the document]
erikhaw: It was supposed to be a little workaround for the test directory, I think I can drop it and handle it in a better way.

quiz/models.py
erikhaw: Sure, this is based on the fact that the quiz template can still change during being assigned, and the max points are calculated based on the questions. Once the quiz is enrolled, the template for it becomes immutable, so I computed max points and stored them. But looking at it now, it would be more logical and optimal to store it in a `TemplateQuiz`, I will change it.

quiz/models.py
```diff
@@ -0,0 +1,75 @@
...
+    # relative quiz path of quiz root directory
+    src = models.CharField(max_length=255, verbose_name="Directory", unique=True)
+    subject = models.ForeignKey(Subject, on_delete=models.CASCADE)
+    semester = models.ForeignKey(Semester, on_delete=models.CASCADE)
```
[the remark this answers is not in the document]
Kobzol: The `Quizzes` page should be independent of assigned quizzes. We can still filter based on the semester mapping of assigned quizzes, e.g. find me all quizzes that have an assignment in semester X.

quiz/settings.py
```diff
@@ -1 +1 @@
-QUIZZ_PATH = "quizzes"
+QUIZ_PATH = "quizzes"
```
Kobzol, line 1: I think that we should directly append this to the base directory, to avoid it being created anywhere working directory is currently set. In theory, the code working with this variable could do it, but I think that it's better if we do it in one place.
```python
import os
from kelvin.settings import BASE_DIR

QUIZ_PATH = os.path.join(BASE_DIR, "quizzes")
```

Kobzol (approved):
> # Model that represents an enrolled quiz

These comments aren't super useful :laughing: We should explain what is the model used for and how do the individual kinds of quizzes differ from each other. In fact it wouldn't be a bad idea to add some README.md to the quiz app to explain the high-level structure of the DB and how does it interact with each other, because it's a bit complex.

But you can do that in another PR. I don't want to block you, so I'm mergin this. Thanks!

PR #627, opened 2025-02-26
Make logging configuration consistent and remove file logging

Kobzol:
Before, we were using a weird combination of the default Django settings and some overridden values.

Now, all logs are written to the console by default, and the format of Django and app logs is unified.

### DIFF
--- a/kelvin/settings.py
+++ b/kelvin/settings.py
@@ -222,40 +222,37 @@ LOGGING = {
     "disable_existing_loggers": False,
     "formatters": {
         "verbose": {
-            "format": "{asctime} - {levelname} - {message}",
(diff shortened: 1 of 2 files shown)


kelvin/settings.py
```diff
@@ -222,40 +222,37 @@
...
-        "user_logins": {  # Custom logger for user login events
-            "handlers": ["file"],
-            "level": "INFO",
+        # Override format of default Django logs
+        "django.server": {
+            "handlers": ["console_http"],
+            "level": "DEBUG",
             "propagate": False,
         },
+        # Disable logs from the serde crate, to avoid spam
```
geordi, line 254: `crate` is a bit Rusty :)

PR #630, opened 2025-02-27
Add a page with student table

Kobzol:
Similar to the already existing task table. This is a prerequisite for adding a student information page (the student page will be searched through this table), which is a prerequisite for displaying student access log information.

CC @patrick11514


geordi (approved):
LGTM and works!

PR #634, opened 2025-03-04
Base components and business logic for quizzes

erikhaw:
Added base components and business logic for quizzes. With this, it's possible to:

From teacher pov:

- CRUD operations on quiz model
- Edit content of quiz
- View quiz
- Assign quiz
- Score quiz

From student pov:

- Enroll quiz
- View quiz results

It's still not "integrated" into Kelvin UI, it's required to call endpoints directly. 

I apologize that this one is quite large, probably largest, but I believe it will not be problematic to read. The next ones will be divisible better.

### DIFF
--- a/api/views/quiz.py
+++ b/api/views/quiz.py
@@ -1,0 +1,319 @@
+import errno
+import json
+import os
+import re
+from shutil import copytree, rmtree, ignore_patterns
+
+from django.utils import timezone
+from serde.json import from_json, to_json
+from serde.yaml import to_yaml
+from unidecode import unidecode
+from django.contrib.auth.decorators import login_required
+from django.contrib.auth.decorators import user_passes_test
+from django.contrib.auth.models import User
+from django.http import (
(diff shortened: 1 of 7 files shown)


api/views/quiz.py
```diff
@@ -0,0 +1,319 @@
...
+    postfix = 1
+
+    # Finding a working directory for new quiz, working directory is considered found if generated path doesn't point to
+    # existing directory and there is no quiz that have generated path assigned.
+    while True:
+        # Assign postfix _X to potential new working directory, where X is natural number
+        postfix_dir_name = dir_name + "_" + str(postfix)
+        postfix_src = os.path.join(src, postfix_dir_name)
+
+        count = Quiz.objects.filter(src=postfix_src).count()
```
Kobzol, line 257: This can race, but I don't think it's super important.

Kobzol:
Oh, this needs a rebase before it can be merged. While at it, please also squash the commits.

Issue #578, opened 2025-01-22
Add types of tasks


geordi:
This will distinguish between homework, real-time test, project, etc.

PR #636, opened 2025-03-16
Task types

Dym03:
Added Task types for possible sorting and grouping of the tasks. Enums are solved using a  [django_enums](https://django-enum.readthedocs.io/en/stable/), could lead to easier comparison of enums.

Fixes: #578


geordi:
Thanks for your contribution! Please, try to squash a few commits.

Kobzol:
For review it's easier to see the individual changes, so you can keep it like this for now. I'll try to take a look soon-ish. Thanks!

common/models.py
```diff
@@ -82,6 +83,14 @@ class Task(models.Model):
     plagiarism_key = models.CharField(max_length=255, null=True)
     created_at = models.DateTimeField(auto_now_add=True)
 
+    class TaskType(models.TextChoices):
+        # Add wanted task types here
+        HOMEWORK = ("HW", "Homework")
```
Kobzol, line 88: Could you please make the field nullable and make the default `NULL`? We have 3k+ existing tasks and I want to distinguish tasks that don't have the type set yet, rather than setting a default for all tasks, which will be incorrect for some of them.

Kobzol: Also, we don't need the (slightly confusing) shortcuts in the DB, let's just use "homework", "test", "project" :)

api/views.py
```diff
@@ -509,6 +509,9 @@ def set_subject(task):
         else:
             task.plagiarism_key = plagiarism_key[:255]
 
+        if data.get("type"):
```
Kobzol, line 512: If the type will be invalid, this will end in 500 server error. Let's check if the type is valid based on the `TextChoices` values, and if not, return a proper 400 bad client request error.

api/views.py
```diff
@@ -597,6 +597,8 @@ def set_subject(task):
         "task_link": reverse("teacher_task", kwargs=dict(task_id=task.pk)),
         "plagcheck_link": reverse("teacher_task_plagiarism", kwargs=dict(task_id=task.pk)),
         "plagiarism_key": task.plagiarism_key,
+        "type": task.type,
+        "task_types": [(item[0], item[1]) for item in Task.TaskType.choices],
```
Kobzol: Let's just hardcode the list in the frontend, it's a bit weird to send a list of possible types with the task data. The frontend should also decide how to render the names of the task types, independently on the label of the DB 
value. (even if the string will be the same for now)

frontend/src/EditTask.svelte
```diff
@@ -328,6 +328,15 @@ async function deleteTask(proceed) {
                           </button>
                         </div>
                       </div>
+                      <div class="col-md">
```
Kobzol, line 331: I think that it doesn't make much sense to say that task A is a homework for students on Monday and a test for students on Tuesday. This is a global property of the task, let's set it outside the class assignments. Also we should handle the NULL case :)

Kobzol (commented):
Thanks a lot for the PR! It looks really good, and would already provide value to us. I left some comments on things that I'd like to change.

Dym03:
I have hopefully resolved all of the problems mentioned in the feedbacks. Only problem with the null values are in the admin page, where they are now required, and i wasn't able to add an option with null as an option, this could be solved with adding a ```blank=True``` in the definition, whereas this would provide null as a default option. This depends on you.  
Small frontend changes were made, preview can be seen on the next picture. 
image

Kobzol:
Sorry, that was a small misunderstanding, we didn't necessarily want to have all the types from Edison, just wanted to point out which ones exist :)

I would go for this set, to make it simpler than Edison:
```
HOMEWORK
PROJECT
EXAM
LABORATORY
OTHER
```

I played with this locally a bit and I have to say that I'm a bit worried about the usage of the enum. It's quite strict and if we ever rename one of the fields or somehow wrong stuff appears in the DB, the server will not have a good time. We could work around that using `type = EnumField(TaskType, null=True, constrained=False, strict=False)`, but at that point we might just as well use a string. This kind of usage also has a big disadvantage in that it changes the field type - if there is an unknown value, it will be a string, while if there's a known value, it will be an enum instance, which is super confusing and error-prone.

I would just use a normal string, as we do for the existing enums (e.g. Day). With `TextChoices`, it's relatively easy to work with them even in the admin. It cannot be reset to `NULL` in admin, but we don't really ever want to reset the type to NULL after setting it for the first time anyway.

Dym03:
Okey, it was not specified, so i added all, but got it, will update that. About the enum thing, i didn't test it that much, but it was only proposed solution. After all the Enum is to my understanding just a wrapper around the choices, so i will change it.

Kobzol:
The enum was a good idea, but I worry that with our state of the database in Kelvin it would be a bit too easy to get into some weird state, and then the website would essentially become bricked until someone updated the DB manually to fix the invalid enum values :sweat_smile:

We don't really create tasks in the admin, so it's not super important how it behaves there. For creatign tasks on Kelvin, on the EditTask page, I would like to uphold the following rules:
- If a task is being created, the task type has to be set (the default should be prefilled e.g. to "homework")
- If a task is being edited, and the the task type was `NULL`, it can remain `NULL` when the task is saved
- If a task is being edited, and the task type was not `NULL`, it shouldn't be possible to set it back to `NULL` again

Does that make sense?

Dym03:
Makes sense. Will try to implement this behavior :). Thanks for the clarification. I wasn't sure where the task where being created, not that is clear.

Dym03:
Should be ready for merge now, squashed down into two main commits for backend and frontend.

frontend/src/EditTask.svelte
```diff
@@ -361,9 +362,27 @@ async function deleteTask(proceed) {
           {/if}
         </div>
 
-        <div title="All tasks with the same plagiarism key will be checked together">
-          <label for="plagiarism-key">Plagiarism key:</label>
-          <input type="text" id="plagiarism-key" maxlength="255" bind:value={task.plagiarism_key} />
+        <div
+          class="input-group mb-3"
+          title="All tasks with the same plagiarism key will be checked together">
```
Kobzol, line 367: This `title` is now applied to both input forms, which doesn't really make sense. Could you please attach it only to the plagiarism input?

I tried wrapping the input and the span in a div and attaching the title to it, but that breaks the layout. If you can figure out how to attach the title to both of them without breaking the layout, please do :) If it's not trivial, then just add the title to the span.

frontend/src/EditTask.svelte
```diff
@@ -361,9 +362,27 @@ async function deleteTask(proceed) {
...
+          <span class="input-group-text" id="basic-addon1">Plagiarism key:</span>
+          <input
+            class="form-control"
+            type="text"
+            id="plagiarism-key"
+            maxlength="255"
+            bind:value={task.plagiarism_key} />
+          <span class="input-group-text" id="basic-addon1">Task type:</span>
```
Kobzol, line 375: The id here is duplicated with the one above. I think that we can remove the ids, also the `plagiarism-key` and `task_type` ids, they don't seem to be used anywhere.

Kobzol (commented):
Thank you! Left two minor comments, otherwise it looks pretty good :)

frontend/src/EditTask.svelte
Dym03: Solved it using a grid :)

Kobzol (approved):
Awesome work. Thank you very much! :heart:

Issue #637, opened 2025-03-17
Edit box in task editor does not list subdirectories


geordi:
Edit box with path in task editor used to display subdirectories when there were some. At this point, it doesn't show anything probably due to the following error in `/api/task-list` endpoint:

```
Internal Server Error: /api/task-list
Traceback (most recent call last):
  File "/home/geordi/projects/kelvin/.venv/lib/python3.12/site-packages/django/core/handlers/exception.py", line 55, in inner
    response = get_response(request)
               ^^^^^^^^^^^^^^^^^^^^^
  File "/home/geordi/projects/kelvin/.venv/lib/python3.12/site-packages/django/core/handlers/base.py", line 197, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/geordi/projects/kelvin/.venv/lib/python3.12/site-packages/django/contrib/auth/decorators.py", line 23, in _wrapper_view
    return view_func(request, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/geordi/projects/kelvin/api/views.py", line 115, in tasks_list_all
    tasks = params.apply(tasks)
            ^^^^^^^^^^^^^^^^^^^
  File "/home/geordi/projects/kelvin/api/views.py", line 84, in apply
    query = query.order_by(*order)
            ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/geordi/projects/kelvin/.venv/lib/python3.12/site-packages/django/db/models/query.py", line 1659, in order_by
    obj.query.add_ordering(*field_names)
  File "/home/geordi/projects/kelvin/.venv/lib/python3.12/site-packages/django/db/models/sql/query.py", line 2231, in add_ordering
    raise FieldError("Invalid order_by arguments: %s" % errors)
django.core.exceptions.FieldError: Invalid order_by arguments: [None]
```

Kobzol:
I noticed this but was too lazy to debug and file and issues, so thanks 😆 I will take a look at this, it's probable that recent changes in the endpoint broke this.

geordi:
Fixed in #638

PR #638, opened 2025-03-18
Fix missing `order_by` in task and student list

Kobzol:
So, the good news is that this was easy to fix. The bad news is that it was broken even before we broke it recently =D The autocomplete tries to download all tasks from Kelvin (there are more than 3 thousand tasks at the moment), which is really not a good idea, and will not work.. It either needs to be reimplemented or we should remove it completely.

At the very least, it should only download tasks from the given subject. I will do that in a follow-up PR.

Related issue: https://github.com/mrlvsb/kelvin/issues/637

### DIFF
--- a/api/views.py
+++ b/api/views.py
@@ -50,11 +50,14 @@ class SearchParams:
     count: int = None
     start: Optional[int] = None
     order_by: Optional[str] = None
-    sort: str = "asc"
+    sort: str = "desc"
 
     @staticmethod
     def from_request(
-        request: HttpRequest, max_count: int, allowed_sort_columns: List["str"]
+        request: HttpRequest,
+        max_count: int,
+        allowed_order_by_columns: List[str],
+        default_order_by: Optional[str],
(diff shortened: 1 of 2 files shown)


api/views.py
```diff
@@ -50,11 +50,14 @@ class SearchParams:
...
     order_by: Optional[str] = None
-    sort: str = "asc"
+    sort: str = "desc"
 
     @staticmethod
     def from_request(
-        request: HttpRequest, max_count: int, allowed_sort_columns: List["str"]
+        request: HttpRequest,
+        max_count: int,
+        allowed_order_by_columns: List["str"],
```
geordi, line 59: Can we get rid of `"` in `"str"`?

geordi (approved):
LGTM!

api/views.py
Kobzol: Yes, thanks! I used literals there before, so it's a leftover from that, probably. Fixed.

PR #640, opened 2025-03-20
Add basic student page

Kobzol:
It doesn't really contain anything useful yet, but I wanted to get the infrastructure in, so that adding stuff to the page can be done with follow-up PRs.

### DIFF
--- a/web/views/teacher.py
+++ b/web/views/teacher.py
@@ -324,10 +324,11 @@ def student_list(request):
 
 
 @user_passes_test(is_teacher)
-def student_page(request, login: int):
+def student_page(request, login: str):
     """
     Page that renders a Vue component with a detail of a single student.
     """
+    login = login.upper()
     student = get_object_or_404(User, username=login)
     if is_teacher(student):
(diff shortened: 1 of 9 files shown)


web/views/teacher.py
```diff
@@ -322,6 +323,22 @@ def student_list(request):
     return render(request, "web/student_list.html")
 
 
+@user_passes_test(is_teacher)
+def student_page(request, login: int):
```
geordi, line 327: I guess, `login` should be `str` not `int`.

web/views/teacher.py
```diff
@@ -322,6 +323,22 @@ def student_list(request):
...
+    """
+    Page that renders a Vue component with a detail of a single student.
+    """
+    student = get_object_or_404(User, username=login)
```
geordi, line 331: In the case I enter it manually in lowercase.

```suggestion
    login = login.upper()
    student = get_object_or_404(User, username=login)
```

Kobzol:
Thanks!

PR #641, opened 2025-03-20
fix (java-pipeline): fix missing parser for unit test ended as error …

david-jezek:
…not only failure

- add /.m2 folder for java pipeline docker image to avoid problem with missing home for some user and running maven
- fix entry script for java pipeline docker image to parse result for unit test ended with result error not only with result failure
- change style for unit test result view in result page to limit height for long stack traces and std outputs


Kobzol (approved):
I won't pretend to understand the Java changes, but in general looks fine :) Thank you!

PR #642, opened 2025-03-20
doc (setup): update run setup description

david-jezek:


docs/setup.md
```diff
@@ -17,7 +17,12 @@ which manages the dependencies of this project. As a first step for working
 with the Python code, install `uv` using some supported approach, for example:
 
 ```bash
-$ pip install --user uv==0.4.4
+$ # Create virtual environment with name venv in folder .venv
+$ python3 -m venv .venv
```
Kobzol, line 21: Hmm, this is not really how `uv` is supposed to be used, it is a global tool that should be installed once per user, and it then manages the virtual environments, libraries, etc. for multiple Python projects.

At the same time, installing `uv` with pip really isn't ideal, in general. I think that we should just provide a link to https://docs.astral.sh/uv/getting-started/installation/.

david-jezek: Ok, I do not know, but when I try to install it in my Ubuntu it always end with even with sudo:

```
jez04@pcfeia406c:~$ sudo pip install --user uv==0.4.4
error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to
    install.
    
    If you wish to install a non-Debian-packaged Python package,
    create a virtual environment using python3 -m venv path/to/venv.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
    sure you have python3-full installed.
    
    If you wish to install a non-Debian packaged Python application,
    it may be easiest to use pipx install xyz, which will manage a
    virtual environment for you. Make sure you have pipx installed.
    
    See /usr/share/doc/python3.12/README.venv for more information.

note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
hint: See PEP 668 for the detailed specification.
jez04@pcfeia406c:~$ 
```

And what I search over an internet I get hint to install it in venv. And its work. May be its problem only in Ubuntu.

Kobzol: Sorry, the documentation was a bit outdated, because before we were combining `uv` with `pip` for legacy reasons, but now we can just use `uv` for everything. I pushed a commit that should hopefully make this clearer.

geordi:
Can we merge it now?

Kobzol (approved):

PR #644, opened 2025-03-20
Allow blank plagiarism key

Kobzol:
This allows us to modify task in the admin page without actually filling in plagiarism key.


geordi (approved):
Works for me!

PR #646, opened 2025-03-20
Add validation to prevent multiple class assignment

### LINKED ISSUE #445: Do not allow to add student to class if he/she is already assigned to a different class in the subject
We're still having this issue that someone adds a student because he/she changed schedule. This leads to inconsistency and student see more assignments that needed.

Dym03:
Fixes #445
Updated validation before adding user to a class to prevent one student having multiple classes from same subject.
Updated errors returned to frontend. Split returned users into 2 groups for easier handling.


Kobzol (commented):
Thanks for the PR! I'm not sure if we can do this before we have class types (similarly as to the recently added task types). Because right now it's quite common to have a student in multiple classes, because there can be e.g. lecture + laboratory class for the same subject, or exam + laboratory classes for the same subject.


@geordi?

Dym03:
That is true, will leave it here, if the class types are added, then it could be finished :). Just to leave the frontend in a good state is there any way to get the inbus import working on a local machine ? Because right now i have a bulk import in a loading state, maybe due to wrong semester_id.

geordi:
At this point, we can give out INBUS API keys. Sorry for that.

PR #647, opened 2025-03-20
Add basic event log

Kobzol:
Adds a table for storing events in the DB, and creates two event actions - login and submit.


geordi (approved):
Works for me!

common/event_log.py
```diff
@@ -0,0 +1,102 @@
...
+    user = models.ForeignKey(settings.AUTH_USER_MODEL, on_delete=models.CASCADE)
+    action = models.CharField(max_length=10, choices=Action.choices)
+    ip_address = models.GenericIPAddressField(null=True, verbose_name="IP address")
+    # Arbitrary metadata that can be attached to each event
+    metadata = JSONField("Metadata", null=True)
+
+    created_at = models.DateTimeField(auto_now_add=True)
+
+    def __str__(self):
+        return f"{self.action} ({self.user.username}) at {self.created_at.strftime("%d. %m. %y %H:%M:%S")}"
```
geordi, line 66: Can we add this info, so it's easier to see addresses in Admin page for now, please?

```suggestion
        return f"{self.action} ({self.user.username}) at {self.created_at.strftime("%d. %m. %y %H:%M:%S")} from {self.ip_address}"
```

geordi (approved):
Works for me!

common/event_log.py
Kobzol: Added.

PR #648, opened 2025-03-21
Display submits in student's page

geordi:


Kobzol (approved):
Hmm, I was thinking of showing the submits directly on this page, but actually just including a link is much easier and we can reuse the existing page. That's a good idea.

geordi:
Yeah, I had the same idea, but we can iterate on that later on and display something like ~20 latest submits or so.

> Hmm, I was thinking of showing the submits directly on this page, but actually just including a link is much easier and we can reuse the existing page. That's a good idea.

PR #655, opened 2025-03-24
Increase nginx file upload size limit to 100 MB

geordi:


deploy/nginx.conf
```diff
@@ -28,6 +28,8 @@ http {
             include /etc/nginx/mime.types;
         }
 
+        client_max_body_size 100M;
```
Kobzol, line 31: Maybe add a comment to keep this in sync with `settings.py`?

Kobzol (commented):
Thanks. That was not that hard, after all =D

Kobzol (approved):
Tested on production, seems to work :+1: I would suggest to lower the size both in nginx and `settings.py` to 25 MiB though. Going from 1 MiB to 100 MiB is quite the jump :)

geordi:
I had a few projects in UPR course that were ~75 MB. So I'd like to keep it like this at the moment, if you don't mind. We can run some stats after the semster regarding the submit sizes.

> Tested on production, seems to work 👍 I would suggest to lower the size both in nginx and `settings.py` to 25 MiB though. Going from 1 MiB to 100 MiB is quite the jump :)

deploy/nginx.conf
geordi: Comment added.

Kobzol (approved):

Kobzol:
Ok :)

PR #656, opened 2025-03-27
Add function for getting currently active exams for a student

Kobzol:
I'm not sure what you want to do with the function yet. For example, forbid users from displaying other assignments when at least a single test is active?


Kobzol:
I updated the function, it should do what we want now.

PR #659, opened 2025-03-30
Remove the `Ctrl + P` component

Kobzol:
This endpoint loads a lot of data on *every page load* done by a teacher, and I'm not sure if anyone uses the component anymore. We now have a dedicated students page and it's quite easy to get to it through various links, so I don't think that this quick action panel is needed anymore (it was cool though :) ).


geordi (approved):
LGTM!

PR #661, opened 2025-04-01
Move comments for non-existent files to summary

Kobzol:
In logs I can see that sometimes there is a user generated comment for a file that is not found. I'm not exactly sure how it can happen (maybe the file disappeared from disk or cannot be loaded?), but in any case we should not error out in that case. This hack moves the comment to the summary in that case.


geordi (approved):
LGTM!

geordi:
I encountered the same behavior a few times as well.

PR #662, opened 2025-04-02
Do not allow connections through non-canonical hosts in nginx

Kobzol:
The invalid host exception was super spammy. I wanted to ignore it first, but I realized that denying these requests is a much better solution than them resulting in a 500 server error (this is kinda stupid behavior btw, Django should really return 400 errors for these...).


geordi (approved):
LGTM!

PR #663, opened 2025-04-02
Added filtering of tasks based on a task types

### LINKED ISSUE #600: Posibility to group tasks by type (prefix) in main view
Location: Kelvin main view, where is list of all task and all groups of students for teacher

Possibility to group or sort task based on task name prefix will be help full. Even for export to CSV file. If students get points from home works and points from labs its unclear how many points is for which type of task on end of the semester.

So group tasks and sum points (even for export) base on task name prefix like "hw_task1" or "lab_java01" will be useful.

Dym03:
Fixes #600
Filters tasks by one task type in the main view, it also calculates the points for the type. 
Right now it looks like this
image. It could also be changed to a select, but i was told, that this would be quicker for the teachers.

### DIFF
--- a/api/views/default.py
+++ b/api/views/default.py
@@ -240,6 +240,7 @@ def class_detail_list(request):
                 "deadline": assignment.deadline,
                 "max_points": assignment.max_points,
                 "students": {s["username"]: {"username": s["username"]} for s in students},
+                "task_type": assignment.task.type,
             }
 
             for score in assignedtask_results(
--- a/frontend/src/ClassDetail.svelte
+++ b/frontend/src/ClassDetail.svelte
@@ -9,10 +9,14 @@ import { user } from './global.js';
 import AddStudentsToClass from './AddStudentsToClass.svelte';
 import Markdown from './Markdown.svelte';
 import AssignmentPoints from './AssignmentPoints.svelte';
+import TaskFilter from './TaskFilter.svelte';
+import { task_types } from './taskTypes.js';
 
 export let clazz;
 export let showStudentsList = clazz['students'].length < 50;
+let task_type = null;
 
+let activeTaskList = [];
 let showAddStudents = clazz.students.length == 0;
 
 let reevaluateLoading = false;
@@ -28,8 +32,14 @@ async function reevaluateAssignment(assignment) {
   reevaluateLoading = false;
 }
 
-function studentPoints(clazz, student) {
-  return clazz.assignments
+$: {
+  activeTaskList = clazz.assignments.filter(
+    (as) => task_type === null || as.task_type === task_type
+  );
+}
+
+function studentPoints(activeTaskList, student) {
+  return activeTaskList
     .map((i) => i.students[student.username])
     .filter((result) => result && result.submits != 0 && !isNaN(parseFloat(result.assigned_points)))
     .map((result) => parseFloat(result.assigned_points))
@@ -37,24 +47,20 @@ function studentPoints(clazz, student) {
     .toFixed(2);
 }
 
-function totalTaskPoints(clazz, assignment_index) {
+function totalTaskPoints(activeTaskList, assignment_index) {
   let assignmentPoints = 0;
 
-  for (let i = 0; i < clazz.students.length; i++) {
-    const student = clazz.students[i];
-    if (!isNaN(clazz.assignments[assignment_index].students[student.username].assigned_points)) {
-      assignmentPoints += Math.max(
-        0,
-        clazz.assignments[assignment_index].students[student.username].assigned_points
-      );
+  Object.entries(activeTaskList[assignment_index].students).forEach(([name, info]) => {
+    if (!isNaN(info.assigned_points)) {
+      assignmentPoints += Math.max(0, info.assigned_points);
     }
-  }
+  });
 
   return assignmentPoints;
 }
 
-function createTaskSummary(clazz, assignment_index) {
-  let maxPoints = clazz.assignments[assignment_index].max_points;
+function createTaskSummary(activeTaskList, clazz, assignment_index) {
+  let maxPoints = activeTaskList[assignment_index].max_points;
   if (isNaN(maxPoints)) {
     maxPoints = 0;
   }
@@ -63,14 +69,12 @@ function createTaskSummary(clazz, assignment_index) {
   let assignmentPoints = 0;
   let gradedStudents = 0;
 
-  for (let i = 0; i < clazz.students.length; i++) {
-    const student = clazz.students[i];
-    const points = clazz.assignments[assignment_index].students[student.username].assigned_points;
-    if (!isNaN(points) && points !== null) {
-      assignmentPoints += Math.max(0, points);
+  Object.entries(activeTaskList[assignment_index].students).forEach(([name, info]) => {
+    if (!isNaN(info.points) && info.points !== null) {
+      assignmentPoints += Math.max(0, info.points);
       gradedStudents += 1;
     }
-  }
+  });
 
   let average = 'N/A';
   if (gradedStudents > 0) {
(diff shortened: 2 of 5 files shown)


geordi:
I had exactly the same opinion! Per class filtering is definitely a way to go along the global one.

Also, I'm not sure we want to add new Svelte code, but in this case, we can make an exception thou.

> Hi, thanks for the PR! I haven't looked at the code yet, but when I saw the screenshot, I thought that it might be useful to apply the filter per class, rather than for the whole page. I'm not sure if it isn't overkill though. @geordi WDYT?

Dym03:
Ok, will work on it

Dym03:
Should work now, not sure about the layout, but i think it is usable. It could be made maybe smaller. The None shows all the tasks, so maybe it should say All, but that all depends. And right now it doesnt implement filtering of tasks without a type. Is that required or no ?
image

api/views/default.py
```diff
@@ -218,68 +221,70 @@ def class_detail_list(request):
...
-                ),
-                "csv_link": reverse(
-                    "download_csv_per_assignment", kwargs={"assignment_id": assignment.id}
-                ),
-                "assigned": assignment.assigned,
-                "deadline": assignment.deadline,
-                "max_points": assignment.max_points,
-                "students": {s["username"]: {"username": s["username"]} for s in students},
-            }
+            if task_type is None or assignment.task.type == task_type:
```
Kobzol, line 224: We should filter the data directly in the DB query if the task type is set, rather than fetching everything from the DB and then filtering it in Python.

Kobzol (commented):
Adding new Svelte code is a bit unfortunate, yes. Just checking: I don't suppose you'd like to try to rewrite the ClassDetail/ClassList components in Vue? :joy:

geordi:
I have branch where I'm working on it. We can rewrite in Vue later on if it's acceptable for you.

> Adding new Svelte code is a bit unfortunate, yes. Just checking: I don't suppose you'd like to try to rewrite the ClassDetail/ClassList components in Vue? 😂

api/views/default.py
Dym03: Sure, Will update that, this Is only used if it is the global filter, otherwise i think it would not be wise to query new data if the per class filter is updated. And it would require more fidelling.

Dym03:
Well I havent used Vue or Svelte, but i can look into that. If it would be of help you can merge it when it is in good state. Then i can try to rewrite it. But i think it will take some time, due to make interconnetiom of svelte and Vue, or is it easy, and simple ? If so i could look in to it next week.

Kobzol:
If you don't have experience with Svelte/Vue, it might take some time to get used to it, Vue is a bit different. It's not *that* much code, so I'm fine with merging the changes in Svelte.

api/views/default.py
Kobzol: Yeah for the classes it's fine to filter in the frontend directly.

Dym03:
Okey, i will change the the query in the backend, then it should be mergable, only think with the global filter is that it doesnt change the sum of the points, Is that wanted, or is it ok to work just with the local filtering.

Kobzol:
Well, it shouldn't produce inconsistent data, of course. The points should always correspond to the displayed tasks, regardless of how they were filtered.

Dym03:
Got it, got an idea how to do it now. Yesterday i wasn't able to get it working.

frontend/src/taskTypes.js
```diff
@@ -0,0 +1,8 @@
+export const task_types = [
+    { key: null, value: 'None' },
```
Kobzol, line 2:
```suggestion
    { key: null, value: 'All' },
```
None is misleading, it sounds like no tasks should be shown.

frontend/src/ClassDetail.svelte
```diff
@@ -82,6 +85,26 @@ Total points: ${assignmentPoints.toFixed(2)}/${totalMaximumPoints}
 Average points: ${average}`;
 }
 
+function createTaskTypeSummary(clazz) {
```
Kobzol, line 88: This would be quite error-prone (as can be seen by the wrong per-student point totals). I would rather have a single function that returns the currently active tasks (after filtering), and then render the table and compute all points and statistics based on the active task list.

Kobzol (commented):
After playing with it locally a bit, I think that maybe we should keep only the per-class filters. It seems too confusing to have both global and local filters, and as you said, the semantics of how it should behave is unclear. (@geordi let me know what do yo think).

When the per-class filters are changed, the student points aren't updated, only the total point counts.

Kobzol:
That sounds like it could work, but I'm not sure if it's worth it to have the global filter in the first place. It is also a bit out of place, because we already have a class filter in the right, so I would perhaps rather add the task type as another (fifth) filter to the class filter instead.

Regarding the points: what I meant is that if we will have to check the filtered task type on multiple places in `ClassDetail`, then it will be very easy to make a mistake and forget to check it somewhere (such as right now the student points displayed on each row were not updated to only consider the filtered tasks). It would be better to implement a single source of truth, i.e. a function/Svelte variable that will return the currently filtered/active tasks for a given class. And then change `ClassDetail` to use the results of this function/Svelte variable, rather than the raw input task array.

Dym03:
Hopefully resolved all the mistakes, from the feedback earlier. If you find something please let me know. The commits could be squashed, when ready to merge, but kept it separate, for the review.

frontend/src/ClassDetail.svelte
```diff
@@ -28,60 +32,76 @@ async function reevaluateAssignment(assignment) {
...
-function createTaskSummary(clazz, assignment_index) {
-  let maxPoints = clazz.assignments[assignment_index].max_points;
+function createTaskSummary(activeTaskList, assignment_index) {
+  let maxPoints = activeTaskList[assignment_index].max_points;
   if (isNaN(maxPoints)) {
     maxPoints = 0;
   }
-
-  let totalMaximumPoints = maxPoints * clazz.students.length;
+  let student_count = Object.entries(activeTaskList[assignment_index].students).length;
```
Kobzol, line 65: It shouldn't be possible to have `activeTaskList[assignment_index].students !== clazz.students`, I think, as tasks are always assigned to the whole class. So I think that the `student_count` can be kept as `clazz.students.length`.

Kobzol (commented):
Thank you, I think that we can be now much more confident that it works.

After trying it, I think that I'm going to throw an annoying business rule into the new beautifully uniform code :laughing: When you filter e.g. only homeworks, the tasks are renumbered (`#1`, `#2`) based on the filtered index. I think that's a bit confusing, as that index is essentially used to identify the individual tasks, if you have the compressed task view enabled (it is controlled by the eye icon on the right of the class detail).

So I would prefer if we showed the task index (`#<T>`) based on the task position within the *unfiltered* task list. Could you try that please? :)

api/views/default.py
```diff
@@ -191,7 +191,8 @@ def all_classes(request):
 @user_passes_test(is_teacher)
```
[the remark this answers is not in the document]
Dym03: Yes, i forgot about it yesterday.

frontend/src/ClassDetail.svelte
Dym03: Sure, it was originaly there.

frontend/src/ClassList.svelte
```diff
@@ -23,9 +23,11 @@ onMount(async () => {
 
 let prevParams;
 async function refetch() {
-  const params = new URLSearchParams(
+  let params = new URLSearchParams(
     Object.fromEntries(Object.entries(filter).filter(([_, v]) => v))
-  ).toString();
+  );
+
+  params = params.toString();
```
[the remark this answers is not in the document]
Dym03: This will be removed aswell, good point. This was updated with the need of global filtering.

Dym03:
With my small testing database, i dont see these small things, as you daily users. So sure i dont think it will be a problem to keep the task index. I will just require to bring back the filtering in the for loop for the task indexes. Will take a look at it asap.

Dym03:
Should be ready to merge now :).

api/views/default.py
```diff
@@ -240,6 +240,7 @@ def class_detail_list(request):
                 "deadline": assignment.deadline,
                 "max_points": assignment.max_points,
                 "students": {s["username"]: {"username": s["username"]} for s in students},
+                "task_type": "" if assignment.task.type is None else assignment.task.type,
```
Kobzol, line 243: `""` should be converted to JSON `null` and handled the normal way in JavaScript, right? I don't think we need to introduce another special empty value. Or was there some special reason why use the empty string?

Kobzol (commented):
Thank you, tried it locally. Have one last question, otherwise it looks great.

Kobzol (approved):
Thanks! Great job :)

PR #664, opened 2025-04-03
Hide nginx version

duy9403:
Exposed server information can lead attackers to find version-specific server vulnerabilities that can be used to exploit unpatched servers.

Untitled

### DIFF
--- a/deploy/nginx.conf
+++ b/deploy/nginx.conf
@@ -3,6 +3,8 @@ events {
 }
 
 http {
+    server_tokens off;
+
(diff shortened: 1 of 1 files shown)


duy9403:
@geordi Please review =) Thank you.

Kobzol (approved):
I think that Kelvin has much worse security problems than exposing the nginx version xD But this change is small enough, so fine by me :)

geordi:
Thanks @duy9403 for the contribution!

PR #666, opened 2025-04-04
Handle move of non-existent files

Kobzol:
This is a bandaid rather than a proper fix, but the frontend FS implementation looks scary, and I got annoyed with seeing this in the logs all the time.


geordi (approved):
Oh nice!

PR #667, opened 2025-04-04
Fix a few npm packages vulnerabilities

geordi:


Kobzol (approved):
dompurify still seems to work, so LGTM.

PR #670, opened 2025-04-10
Add relative deadline button

### LINKED ISSUE #551: Allow copying relative deadline
When there are many classes, it's super annoying to manually set deadlines for a task. When the deadline is the same, there is already a copy button, but if the deadlines differ (e.g. for a week-long test), it's quite cumbersome to fill it everywhere. It would be nice if there was a way to copy a "relative deadline", i.e. I configure the start and end of the first class, and then I copy the same relative duration of the task for all remaining classes.

Dym03:
Fixes #551
This is the way that I understood relative deadline. You specify start and end in the first class, and the just specify the start of each of the other classes and then just copies the relative deadline. So the deadline would be Assigned + (diff from the one that it was clicked from).

### DIFF
--- a/frontend/src/EditTask.svelte
+++ b/frontend/src/EditTask.svelte
@@ -193,6 +193,16 @@ function assignSameToAll(templateClass) {
   });
 }
 
+function setRelativeDeadlineToAssigned(assigned, deadline) {
+  const diff = new Date(deadline) - new Date(assigned);
+  task.classes = task.classes.map((cl) => {
+    if (cl.assigned) {
+      cl.deadline = new Date(new Date(cl.assigned).getTime() + diff);
+    }
+    return cl;
+  });
+}
+
 async function duplicateTask() {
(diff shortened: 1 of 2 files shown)


geordi:
Thank you for the contributions!

Please, make the fix in a separate PR, even it's a very small change.

I'll take a look to the relative button feature ASAP.

Kobzol (approved):
I really don't want to add new Svelte code :( But the functionality is great, thank you :)

Dym03:
I took a quick look into Vue. But the main page is so dependent on other svelte code. That i though of implementing the functionality. And if i find the time, i would try to implement the Vue counterparts. But i think it will take some time.

Kobzol:
Yeah, it wasn't meant as a complaint, it's not easy to move the task detail page to Vue :) @patrick11514 just had a whole bachelor thesis on this, where he ported a few other pages and wrote new Vue components.

PR #671, opened 2025-04-10
Added hard deadline option

### LINKED ISSUE #567: Add hard deadline to tasks
We should have the option to set a "hard deadline", where Kelvin won't allow students to submit anything after the deadline has passed.

Dym03:
Fixes #567

It only shows up if the deadline is assigned. When the student sends submit after the deadline, it stops the submit and returns a error message. Might be a good thing, to maybe have some king of signalization about the hard deadline for the students. The same about the teachers. But i will leave that up to you. 
image


geordi:
Thanks! I'll take a look into that.

frontend/src/EditTask.svelte
```diff
@@ -313,6 +321,28 @@ async function deleteTask(proceed) {
                         semesterBeginDate={$semester.begin}
                         onFromDuplicateClick={setAssignedDateToVisible}
                         onToDuplicateClick={setDeadlineToAssigned} />
+                      {#if clazz.deadline}
+                        <div class="col-md">
+                          <div class="input-group">
+                            <input
```
geordi, line 327: Would it be possible to change the checkbox button to a more [classic](https://getbootstrap.com/docs/5.0/forms/checks-radios/#checks) style? I somehow struggle to immediately see if it's checked or not.

api/views/default.py
```diff
@@ -911,6 +913,13 @@ def create_submit(request: django.http.HttpRequest, task_assignment: int) -> Jso
             },
             status=400,
         )
+    except SubmitPastHardDeadline:
+        return JsonResponse(
+            {
+                "error": "The submit was send after a deadline, that was set as final, so it is not valid submit"
```
geordi, line 919: Please, reformulate the error message to: The submission was sent after the deadline, which was final, so it is not a valid submission.

Including the trailing "dot" :-)

geordi (commented):
In overall it's nice contribution. Thank you!

Dym03:
Fixed the error message, and updated the checkbox. 
Current state:
image
image

geordi (approved):
LGTM!

common/models.py
```diff
@@ -226,6 +226,7 @@ class AssignedTask(models.Model):
     clazz = models.ForeignKey(Class, on_delete=models.CASCADE)
     assigned = models.DateTimeField()
     deadline = models.DateTimeField(null=True, blank=True)
+    hard_deadline = models.BooleanField(null=True, blank=True)
```
Kobzol, line 229: Let's make this `NOT NULL`, and make the `db_default` be `False`, to backfill it into the old rows when the migration is performed (we should check that this works fine).

In theory we could make it an enum/text to be more forward-compatible, but I don't have a use-case for any other deadline kind, so let's keep the bool for now, we can always change it in th efuture.

api/views/default.py
```diff
@@ -911,6 +913,13 @@ def create_submit(request: django.http.HttpRequest, task_assignment: int) -> Jso
             },
             status=400,
         )
+    except SubmitPastHardDeadline:
```
Kobzol, line 916: This is the JSON API submit interface, but we also need to explicitly catch this error in `task_detail`, otherwise it produces a 500 server error.

api/views/default.py
```diff
@@ -633,6 +633,7 @@ def set_subject(task):
                         if cl.get("deadline", None)
                         else None,
                         "max_points": cl.get("max_points", None),
+                        "hard_deadline": cl.get("hard_deadline", None),
```
Kobzol, line 636: When you set a deadline, and then set a hard deadline, and then you unset the deadline, the DB will contain `deadline=NULL` and `hard_deadline=True`. We need to figure out what to do about this.

Either we allow storing `NULL` in `hard_deadline` (as you do now) and reset `hard_deadline` to `NULL` when `deadline` is not set. Or we can keep the original value of `hard_deadline` in the DB, but the check in `store_submit` has to assume that `hard_deadline` might be set even if there is no deadline (currently it doesn't do that).

I would prefer the latter, and add a method to `AssignedTask` (`has_hard_deadline()`) that will return `True` only if `hard_deadline` is `True` and `deadline` is not NULL.

common/submit.py
```diff
@@ -33,6 +38,11 @@ def store_submit(request: HttpRequest, assignment: AssignedTask) -> Submit:
     Optionally, the request can also contain a "paths" field that contains a string with a single
     path per line, assigning paths to the uploaded "solution" files.
     """
+    if (
```
Kobzol, line 41: This check has to be modified (see other comment), in case deadline is `NULL`, but `hard_deadline` is set.

frontend/src/EditTask.svelte
```diff
@@ -324,6 +332,26 @@ async function deleteTask(proceed) {
...
+                              type="checkbox"
+                              id="hardDeadline_{idx}"
+                              bind:checked={clazz.hard_deadline} />
+                            <label class="form-check-label check_box_label" for="hardDeadline_{idx}"
+                              >Hard Deadline</label>
+                            <button
+                              class="btn btn-sm btn-secondary"
+                              on:click|preventDefault={() =>
+                                assignHardDeadlineToAll(clazz.hard_deadline)}
+                              title="Set all hard deadlines">
```
Kobzol, line 349:
```suggestion
                              title="Copy hard deadline setting to all classes">
```
Otherwise it sounds like it sets (to true) the option for all classes.

frontend/src/EditTask.svelte
```diff
@@ -324,6 +332,26 @@ async function deleteTask(proceed) {
                         onFromDuplicateClick={setAssignedDateToVisible}
                         onToDuplicateClick={setDeadlineToAssigned}
                         onToRelativeClick={setRelativeDeadlineToAssigned} />
+                      {#if clazz.deadline}
+                        <div class="col-md">
```
Kobzol, line 336: Let's add a `title` here to explain what hard deadline does. E.g. `Forbids students to submit solutions after the deadline has passed.`.

Kobzol (commented):
Thanks for the PR! There are some changes that need to be made to it, left some comments.

We should also make the "Submit" button for students disabled when the deadline has passed and it is a hard deadline.

common/models.py
Dym03: Will do that. Well my first idea was something like a soft deadline, that would not enable submits like 5-10 minutes after the deadline that was set. But thats why i asked about the bool. If this would be wanted, sure could be redone to enum.

api/views/default.py
Dym03: Absolutely true, thanks for the point. Will do that.

frontend/src/EditTask.svelte
Dym03: Sounds better, thanks.

frontend/src/EditTask.svelte
Dym03: Sure. Will add it there.

api/views/default.py
Dym03: Will look into that, to resolve it.

common/models.py
Dym03: So db_default cannot be added as of yet. In kelvin django is in v4.2.16 but the db_default was added with [v5.0](https://docs.djangoproject.com/en/5.2/releases/5.0/). I took a quick look at how could this be solved and found this [article](https://www.viget.com/articles/dealing-with-database-defaults-in-django/). Where in the migration a sql script is included that takes care of it. So I propose this setting ```hard_deadline = models.BooleanField(blank=True, null=False, default=False)```. And add the script in the migration.

Kobzol: We don't actually need DEFAULT in the DB. As long as the migration works, it's fine. So if Django adds DEFAULT for the migration itself, it should be fine.

Dym03:
All of the mentioned above should be resolved. Instead of disabling the submit button, I took advantage of the upload flag, that was there. That at least with my testing disables also the option of drag and drop submits. The migration locally inserted the default value to all the rows. So i think it might work. But please test it. Otherwise a sql script would be needed that would run with the migration, that would set he values.

api/views/default.py
```diff
@@ -633,6 +633,7 @@ def set_subject(task):
                         if cl.get("deadline", None)
                         else None,
                         "max_points": cl.get("max_points", None),
+                        "hard_deadline": cl.get("hard_deadline"),
```
Dym03, line 636: As of my understanding, with the updated code, where null=False, it shouldn't be able to have undefined/null values here. So this should be valid. Please correct me if i am wrong.

common/models.py
```diff
@@ -226,13 +227,19 @@ class AssignedTask(models.Model):
     clazz = models.ForeignKey(Class, on_delete=models.CASCADE)
     assigned = models.DateTimeField()
     deadline = models.DateTimeField(null=True, blank=True)
-    hard_deadline = models.BooleanField(null=True, blank=True)
+    hard_deadline = models.BooleanField(blank=True, null=False, default=False)
```
Kobzol, line 230: I think that we can omit the `blank` attribute, it should only affect the admin (because we don't use Django forms), and since it's a boolean, it cannot be "not filled". `null=False` is also the default, so `default=False` should be enough.

frontend/src/EditTask.svelte
```diff
@@ -333,7 +333,9 @@ async function deleteTask(proceed) {
                         onToDuplicateClick={setDeadlineToAssigned}
                         onToRelativeClick={setRelativeDeadlineToAssigned} />
                       {#if clazz.deadline}
-                        <div class="col-md">
+                        <div
+                          class="col-md"
+                          title="Forbids students to make submitions after the deadline has passed.">
```
Kobzol, line 338: `submitions` -> `submissions`

web/views/student.py
```diff
@@ -373,6 +374,11 @@ def task_detail(request, assignment_id, submit_num=None, login=None):
                 f"Too many submits. You need to wait {e.time_until_limit_expires.total_seconds():.0f}s before sending another submit",
                 status=429,
             )
+        except SubmitPastHardDeadline:
+            return HttpResponse(
+                "You have submitted after deadline, which was set as hard, so no submits after it will be accepted.",
```
Kobzol, line 379: Your submit was sent after the deadline, which is not allowed for this task.

Kobzol (commented):
Thanks! Left a few typos/rewordings.

Using `upload: False` is good, but I think that it would be nice to also disable the upload button in the frontend. During a test, students often open the task page in a tab and have it opened for a long time, during which the deadline can expire. So it would be great if we could interactively disable the button (and drag & drop, just disable the upload functionality altogether) directly in the frontend too. But let's leave that for another PR.

Kobzol:
```
python3 manage.py sqlmigrate common 0028_assignedtask_hard_deadline
BEGIN;
--
-- Add field hard_deadline to assignedtask
--
ALTER TABLE "common_assignedtask" ADD COLUMN "hard_deadline" boolean DEFAULT false NOT NULL;
ALTER TABLE "common_assignedtask" ALTER COLUMN "hard_deadline" DROP DEFAULT;
COMMIT;
```
Migration SQL looks good.

Dym03:
Fixed the typos. Also added the disabled button, not sure why it was span at first, had to switch it to button, so hope it doesn't matter. Didn't manage to disable the drag and drop. Not sure how that is handled.

templates/web/task_detail.html
```diff
@@ -266,6 +266,20 @@ <h5>
             window.location.href = keyTabMap[e.code];
         }
     });
+    const button = document.getElementById('upload_button');
```
Kobzol, line 269: We should set the disabled attribute when the page is loaded. Otherwise then the deadline has expired, you load the page, the button is enabled, and then it gets disabled after a second. That's not very good UX :)

templates/web/task_detail.html
```diff
@@ -159,7 +159,7 @@ <h5 title="Top 5 plagiarism matches for some checked submit of this student" cla
                             <input type="file" name="solution" multiple hidden id="upload-choose"
                                    onclick="this.value=null;"
                                    onchange="this.form.submit()">
-                            <span class="btn btn-sm btn-primary" onclick="document.getElementById('upload-choose').click()">Upload</span>
+                            <button id="upload_button" class="btn btn-sm btn-primary" onclick="document.getElementById('upload-choose').click()" data-deadline="{{ deadline|date:'c' }}">Upload</button>
```
Kobzol, line 162: Just using a button doesn't work, if you click on "Upload", it will send an empty request to the server :) So either the button behavior should be fixed, or we should keep `span` and disable it manually (by changing color and disabling the click functionality).

templates/web/task_detail.html
```diff
@@ -266,6 +266,20 @@ <h5>
             window.location.href = keyTabMap[e.code];
         }
     });
+    const button = document.getElementById('upload_button');
+    const deadlineStr = button.getAttribute('data-deadline');
+    const deadline = new Date(deadlineStr);
+    function updateUploadState() {
```
Kobzol, line 272: This doesn't respect the hard deadline configuration :)

templates/web/task_detail.html
Dym03: I might be wrong here. But If the deadline has expired and you load the page, the submit button would not be rendered thanks to the if statement above. But it's true, if the request take longer, then it might end up in a state you said. So i will add the function call at the  onload event. If that is what you meant.

Kobzol: I tried it locally and the button has appeared and then got disabled after a second. But in any case it should be just disabled or hidden immediately when the hard deadline has expired.

Dym03: Yes sorry, there is bug. Currently it would disable the button, even if the deadline wasn't hard. And that is not wanted. I apologize, today must have been a pain for the reviews.

Kobzol:
Thank you, the upload button now seems to behave fine. One thing I noticed now is that when the upload button is hidden, it gives the possibility for teachers to grade themselves on the task detail page xD But more importantly it actually prevents them from submitting code to the tasks, which is a bit annoying, because sometimes it is useful to provide more code for plagiarism detection.

Could you please modify the code to ignore hard deadline for teachers? Both in the submit check function and in task detail. I think that's the last thing I'd like to change here, otherwise it looks great.

Dym03:
Sure will look into that.

web/views/student.py
```diff
@@ -293,12 +293,16 @@ def task_detail(request, assignment_id, submit_num=None, login=None):
...
         "assigned": assignment.assigned if is_announce else None,
         "deadline": assignment.deadline,
+        "hard_deadline": assignment.hard_deadline,
         "submits": submits,
         "text": testset.load_readme().announce if is_announce else testset.load_readme(),
         "inputs": None if is_announce else testset,
         "max_inline_content_bytes": MAX_INLINE_CONTENT_BYTES,
         "has_pipeline": bool(testset.pipeline),
-        "upload": not user_is_teacher or request.user.username == login,
+        "upload": (not user_is_teacher or request.user.username == login)
```
Kobzol, line 302: What about creating a variable `hard_deadline` that will be set to `assignment.hard_deadline and not user_is_teacher`? Then you don't need to have this complex condition for the `upload` vaue, and you don't need to check if the user is a teacher in the frontend at all, the frontend will just see `hard_deadline: false`.

Dym03: True, that would make it simpler. Will update that asap.

Dym03: I did as you said, but that brought up some difficulties, with the feature of not displaying the submit at all, if the hard deadline is over. So i kept it simple, with just he button being disabled for now. I didn't found a way to implement it without the complicated upload variable. Due to the lack of () in django templates.

Kobzol: Maybe we misunderstood each other. I meant to do this:
```python
hard_deadline = assignment.hard_deadline and not user_is_teacher

{
...
"hard_deadline": hard_deadline,
"upload": (not user_is_teacher or request.user.username == login)
        and not (hard_deadline and assignment.is_past_deadline())
}
```
That seems like it should work.

Kobzol (approved):
Thank you! :)

PR #672, opened 2025-04-10
Fixed more content hover in Class Detail

Dym03:
Fixes new bug found in my recenlty added feature.
Before
image
After
image


Kobzol (approved):
Thanks!

PR #673, opened 2025-04-11
Changed white font color to grayish in dark mode

### LINKED ISSUE #573: Font color in dark mode
Would be nice to change font color in dark mode from white to some "grayish" color. Glowing white is especially noticeable in list of students, total points, etc.

Dym03:
Fixes #573

Updated 2 elements, 1 class with the greyish color. The td, th tags i think caused the main problem. The global color change wouldn't change the color of these elements, or atleast with my experiments it didn't.
Current state:
image

### DIFF
--- a/frontend/src/global.scss
+++ b/frontend/src/global.scss
@@ -42,4 +42,9 @@ html[data-bs-theme='dark'] {
     .d2h-file-header {
         background-color: #343434;
     }
+    th,
+    td,
+    .btn-dark {
+        color: var(--bs-body-color);
+    }
 }


Kobzol:
I can't say that I see much difference :laughing: I'll let geordi evaluate the color :)

Dym03:
Sure can be changed more. In RGB it is now rgb(222, 226, 230), compared to 255,255,255. Tried other variants. But this is some kind of global variable, that is used all over the place, so i didn't want to make it look funny with all kinds of different shades of gray.

geordi (approved):
Much better color! :-)

Kobzol:
image

Hmm, this is not exactly ideal xD

Dym03:
Should have some time today, so i will try look into that. Didn't think this was the effect that it had. Thanks for the notice.

Issue #677, opened 2025-04-25
Error in task assignment


geordi:
Assigning of a task (without specifying hard deadline) fails with the following error:

```
null value in column "hard_deadline" of relation "common_assignedtask" violates not-null constraint
DETAIL:  Failing row contains (8849, 2025-04-26 11:30:00+00, 2025-04-26 12:30:00+00, 30, 1105, 3331, null, null).
```

Dym03:
Weird, will take a look at that asap.

Kobzol:
https://github.com/mrlvsb/kelvin/pull/678 should fix this.

PR #678, opened 2025-04-25
Handle missing `hard_deadline` in `task_detail`

Kobzol:
This logic is brutal.. When a new task is being created, the `EditTask` components loads classes from `/api/subject`, and then monkey patches additional attributes to these classes, and then it resends the classes (which then actually represent assigned tasks) to `task_detail`. This should be rewritten from scratch =D

This PR is mostly a hotfix.

Fixes: https://github.com/mrlvsb/kelvin/issues/677


Dym03:
Thanks for the fix, seems like my intuition that default would prevent null from there was false. 🙂

PR #679, opened 2025-05-02
Quiz base frontend

erikhaw:
Added base frontend for quizzes. With this, it's possible to:

From teacher pov:

CRUD operations on quiz model
Edit content of quiz
View quiz
Assign quiz
Score quiz
From student pov:

Enroll quiz
View quiz results

It's still not "integrated" into Kelvin UI, it's required to call endpoints directly.

### DIFF
--- a/api/views/quiz.py
+++ b/api/views/quiz.py
@@ -252,13 +252,13 @@ def quiz_add(request: HttpRequest):
     while True:
         # Assign postfix _X to potential new working directory, where X is natural number
         postfix_dir_name = dir_name + "_" + str(postfix)
-        postfix_src = os.path.join(src, postfix_dir_name)
+        disk_dir_src = os.path.join(src, postfix_dir_name)
 
-        count = Quiz.objects.filter(src=postfix_src).count()
+        count = Quiz.objects.filter(src=os.path.join(abbr, username, postfix_dir_name)).count()
 
         if count == 0:
             try:
-                os.makedirs(postfix_src)
+                os.makedirs(disk_dir_src)
 
                 break
             except OSError as e:
--- a/frontend/src/Quiz/Quiz.vue
+++ b/frontend/src/Quiz/Quiz.vue
@@ -1,0 +1,453 @@
+<script setup lang="ts">
+/**
+ * This component displays a quiz that can "work" in 4 modes, based on passed parameters.
+ * It is available for teachers and students.
+ *
+ * For teacher, it can act like a component to view detail of quiz, or to score quiz.
+ * For student, it can act like a component to fill a quiz, or to view quiz results.
+ */
+import { ref, onMounted, onUnmounted } from 'vue';
+import VueCountdown from '@chenfengyuan/vue-countdown';
+import { getDataWithCSRF } from '../utilities/api';
+
+const { quiz_html, is_teacher, student, enrolled_id, answers, remaining, scoring, scored_by } =
+  defineProps<{
+    quiz_html: string;
+    user: number;
+    student: string;
+    remaining: string;
+    answers: string;
+    scoring: string;
+    quiz_id: number;
+    is_teacher: string;
+    enrolled_id: number;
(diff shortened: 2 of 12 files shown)


frontend/src/main.js
```diff
@@ -193,6 +195,8 @@ registerSuspendedVueComponent('task-list', TaskList);
 registerVueComponent('student-list', StudentList);
 registerSuspendedVueComponent('inbus-import', InbusImport);
 registerVueComponent('notifications', NotificationsNew);
+registerSuspendedVueComponent('quiz', Quiz);
```
Kobzol, line 198: It doesn't look like these components have a top-level `await`, so they don't need to be suspended.

templates/web/quiz/quiz.html
```diff
@@ -0,0 +1,26 @@
...
+            <h2>
+                {{ quiz.title }}
+                {% if is_teacher and scoring is None %}
+                    <a href="/teacher/quiz/{{ quiz.id }}/edit" class="text-muted" title="Edit quiz"><span
+                            class="iconify"
+                            data-icon="fa-solid:pen"></span></a>
+                {% endif %}
+            </h2>
+        </div>
+        <kelvin-quiz quiz_html="{{ quiz_html }}" is_teacher="{{ is_teacher|lower }}"
```
Kobzol, line 19: It's unfortunate that we would have to pass everything as strings because it is a web component. Please take a look at `student_page.html` to see how we can do it properly by passing the typed props to the Vue component, without going through web components.

frontend/src/Quiz/QuizAssign.vue
```diff
@@ -0,0 +1,318 @@
...
+      if (!isMissingPropertyToAssign(clazz)) {
+        clazz.publish_results = true;
+      }
+    });
+};
+
+/**
+ * Watch for changes in the selected teacher and filter the classes
+ */
+watchEffect(() => {
```
Kobzol, line 155: This is a really hacky solution :) Please create a `computed` that will return only the visible classes based on `selectedTeacher`, and use that in the template, to avoid modifying the classes.

frontend/src/Quiz/QuizAssign.vue
```diff
@@ -0,0 +1,318 @@
...
+            </tr>
+          </tbody>
+        </table>
+      </div>
+    </div>
+  </div>
+</template>
+
+<style scoped>
+#assign:deep(td) {
```
Kobzol, line 300: Why is `:deep` here? That is only required if you wanted to style `td` inside a nested Vue component, but I don't see any of that in this template, it's just HTML.

frontend/src/Quiz/Quiz.vue
```diff
@@ -0,0 +1,453 @@
...
+  }>();
+
+enum Modes {
+  Display,
+  Filling,
+  Scoring,
+  ViewScoring
+}
+
+enum ResultsType {
```
Kobzol, line 34: I guess that this represents something like `SubmitType`? And the values should be `Periodic` and `Manual/Final`.

Kobzol (commented):
Thank you for the PR, left a bunch of comments.

Could you please remove the Vue stuff from this PR? It's too much for me :) And it's also difficult to review without actually trying the UI. The Vue UI components are very large and hard to understand, they should be split into a smaller number of components.

api/views/quiz.py
```diff
@@ -317,3 +326,191 @@ def quiz_duplicate(request, quiz_id):
         file.write(str(quiz_copy.pk))
 
     return JsonResponse({"id": quiz_copy.pk})
+
+
+# Function that allows enrolling student to a quiz.
```
geordi, line 331: Function comments should be in `"""` and in the function body.

Kobzol (commented):
Thanks. Please squash the commits to remove the merge commits, and we'll merge it.

Kobzol:
Thank you!

Issue #615, opened 2025-02-17
Automatic deployment from GitHub


Kobzol:
Currently, we deploy Kelvin manually after changes are pushed to it on GitHub. It would be great to figure out a way how to automate that.

PR #680, opened 2025-05-12
Initial Proposal for Automated Deployment from GitHub

JersyJ:
## Issue
Part of #615.

## Summary
Initial proposal for automating the deployment process from GitHub to the Kelvin Server (VM).


Kobzol:
Hi, sorry for the delay, I was at a conference last week. The overall idea looks solid, I was imagining it very similarly, but I think that the most important detail is not described here yet - where will the `/deploy` service be running and how it will communicate with Docker. I suppose that the service can either be directly in Nginx, be a part of Kelvin, or be a separate web service. In any case we need to 1) somehow route web requests through Nginx, and 2) make sure that it can actually invoke Docker commands *from within Docker itself*.

JersyJ:
> Hi, sorry for the delay, I was at a conference last week. The overall idea looks solid, I was imagining it very similarly, but I think that the most important detail is not described here yet - where will the `/deploy` service be running and how it will communicate with Docker. I suppose that the service can either be directly in Nginx, be a part of Kelvin, or be a separate web service. In any case we need to 1) somehow route web requests through Nginx, and 2) make sure that it can actually invoke Docker commands _from within Docker itself_.

Hi, I have updated the docs to reflect this feedback.

- Regarding the /deploy service: it should run as a separate, lightweight Docker container and use the Docker-outside-of-Docker (DooD) approach to communicate with the host Docker daemon. This allows the service to invoke Docker commands from within its own container.

- The service will be exposed through the /deploy route via Nginx. Routing is handled through a straightforward change to the nginx.conf.

JersyJ:
Just sharing my thoughts:

 I know this might seem a bit complex.

I did evaluate a self-hosted runner that uses a simpler pull-based model, but it's not recommended for public repositories due to serious security concerns.

Based on my research, this is the most straightforward way to build a safe and reliable automated deployment with built-in rollback if something fails without the added complexity of platforms like Kubernetes, K3s, or Nomad.

 Let me know if you're okay with this, and I’ll start the implementation.

Kobzol:
Thank you for the workflow diagram, that's really awesome. I'm happy with this proposal and would love to see it go forward. Having a sync deployment mechanism combined with minimal downtime and automatic rollback is really nice.

One thought that I don't was discussed here: if a rollback happens, we'd still keep the `latest` tag in the registry set to the failing version (if we push `latest` from the GHA workflow). I guess we could do it like this:
1) Push image to registry
2) Send it to the deployment service
3) Deploy it, check that it's ok
4) If yes, return 200, and only then the GHA workflow also updates the latest tag in the registry

JersyJ:
> Thank you for the workflow diagram, that's really awesome. I'm happy with this proposal and would love to see it go forward. Having a sync deployment mechanism combined with minimal downtime and automatic rollback is really nice.
> 
> One thought that I don't was discussed here: if a rollback happens, we'd still keep the `latest` tag in the registry set to the failing version (if we push `latest` from the GHA workflow). I guess we could do it like this:
> 
> 1. Push image to registry
> 2. Send it to the deployment service
> 3. Deploy it, check that it's ok
> 4. If yes, return 200, and only then the GHA workflow also updates the latest tag in the registry

Good point, your suggestion makes sense. I’ve updated the docs.

JersyJ:
Should we merge this first, and I submit a separate PR for the implementation? Or would you prefer I continue working in this one?

Kobzol:
We shouldn't merge documentation about something that doesn't exist yet :) So feel free to continue here (or send a different PR, and we'll merge this one once the documentation actually matches reality). It doesn't have to all be in a single PR, of course.

JersyJ:
I believe this PR is ready for review @Kobzol. I’ll handle the GitHub Action changes in a separate PR. I hope the size won`t be problem :D, most of it comes from the 100-character line limit and the tests. 

Btw here is little demo how it works:
https://github.com/user-attachments/assets/8d455a7e-50d7-4414-8ef9-ce435b4da430

Kobzol:
Umm, yeah, this is a big big :sweat_smile: I think it would be best if you could walk me through it.

docker-compose.yml
```diff
@@ -75,6 +75,25 @@ services:
...
+    profiles: [ prod ]
+    build:
+      context: deployment_service
+      dockerfile: Dockerfile
+      target: runtime
+    image: "ghcr.io/mrlvsb/deployment:latest"
+    pull_policy: always
+    restart: unless-stopped
+    environment:
+      - DOCKER__COMPOSE_FILE_PATH=/kelvin/docker-compose.yml
```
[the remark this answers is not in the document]
JersyJ: Fixed

Kobzol:
I'm not sure if that's not too heavyweight for us. Let me know if you want me to do another review pass on the PR.

docker-compose.yml
```diff
@@ -75,6 +75,25 @@ services:
...
+    image: "ghcr.io/mrlvsb/deployment:latest"
+    pull_policy: always
+    restart: unless-stopped
+    environment:
+      - DOCKER__COMPOSE_FILE_PATH=${DOCKER__COMPOSE_FILE_PATH}
+      - SECURITY__WEBHOOK_SECRET=${SECURITY__WEBHOOK_SECRET}
+      - SECURITY__BACKEND_CORS_ORIGINS=${SECURITY__BACKEND_CORS_ORIGINS}
+      - SECURITY__ALLOWED_HOSTS=${SECURITY__ALLOWED_HOSTS}
+    volumes:
+      - ${REPO__DIRECTORY_PATH}:/kelvin
```
Kobzol, line 94: Please remind me what is this for? I haven't found any mention of `/kelvin` in the deployment service.

deployment_service/app/main.py
```diff
@@ -0,0 +1,139 @@
...
+app = FastAPI(
+    title="Deployment Service",
+    description="A Service to do on-premises deployments.",
+)
+
+# Sets all CORS enabled origins
+app.add_middleware(
+    CORSMiddleware,
+    allow_origins=[
+        str(origin).rstrip("/") for origin in get_settings().security.backend_cors_origins
```
Kobzol, line 19: Why do we need CORS at all, when the service won't be accessed by browsers?

deployment_service/app/main.py
```diff
@@ -0,0 +1,139 @@
...
+    ],
+    allow_credentials=True,
+    allow_methods=["*"],
+    allow_headers=["*"],
+)
+
+# Guards against HTTP Host Header attacks
+app.add_middleware(
```
Kobzol, line 27: Same here, we only care about the signed token, not where the request is coming from, right?

deployment_service/app/main.py
```diff
@@ -0,0 +1,139 @@
...
+    TrustedHostMiddleware,
+    allowed_hosts=get_settings().security.allowed_hosts,
```
Kobzol, line 29: If we remove the host & CORS stuff, then we can get rid of `pydantic-settings`, which is IMO overkill for this. We can just load a few environment variables the good old fashioned way, and we don't need this library.

deployment_service/app/deployment.py
```diff
@@ -0,0 +1,313 @@
...
+            self.logger.info(
+                f"Copying candidate docker-compose.yml file at {candidate_dir} for commit {self.commit_sha}"
+            )
+            if not await self._run_command(
+                [
+                    "git",
+                    "show",
+                    f"{self.commit_sha}:{os.path.basename(self.stable_compose_path)}",
+                    ">",
+                    candidate_compose_path,
```
Kobzol, line 255: The command executes bash? We can just pipe the output to the given file path, not need to use `>`.

Kobzol (commented):
I think that the high-level logic makes sense, and we can try to test it on production e.g. over a weekend. But also since it's a lot of code, I would like to remove all the things that are not needed before we merge this. So I'd say get rid of CORS and allowed hosts, which I think are not needed here, which allows us to get rid of the magical environment variable parsing.

I also think that a lot of the test are not very useful, especially those that use mocks heavily, and they are essentially write-only code for Kelvin. But I don't want to go cherry pick what tests to keep, so let's just keep them all.

docker-compose.yml
```diff
@@ -91,7 +91,7 @@ services:
       - SECURITY__BACKEND_CORS_ORIGINS=${SECURITY__BACKEND_CORS_ORIGINS}
       - SECURITY__ALLOWED_HOSTS=${SECURITY__ALLOWED_HOSTS}
     volumes:
-      - ${REPO__DIRECTORY_PATH}:/kelvin
+      - ${REPO__DIRECTORY_PATH}:${REPO__DIRECTORY_PATH_IN_CONTAINER}
```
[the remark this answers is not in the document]
JersyJ: Done, please check if it is enough

JersyJ:
> I set up the deployment on another machine and found a small bug caused by the removal of worktrees. It’s not ready to merge yet, but the fix should be minor. Let me try to fix that quickly.

Ok, the issue wasn’t in the code but in my setup 😄. The deployment now works as expected. However, removing the worktree made the setup less flexible, since you can’t, for example, rely on updated files like `nginx.conf` or other components mounted from the repository. I also removed a few tests that didn’t add any real value.

Kobzol:
> rely on updated files like nginx.conf or other components mounted from the repository.

The only file for deployment that should be taken from the git repo is `docker-compose.yml`, the rest are just example files, so that should actually be a feature, not a bug.

Kobzol (approved):
Thank you for implementing and testing it. I will merge it now and try to run it on production to see what happens.

deployment_service/app/deployment.py
```diff
@@ -213,10 +213,10 @@ async def _health_check(self) -> bool:
         """Performs a health check by making HTTP requests to a specified URL."""
         self.logger.info(f"Performing health check on {self.healthcheck_url}...")
         end_time = time.time() + HEALTH_CHECK_TIMEOUT
-        async with httpx.AsyncClient(timeout=2.0) as client:
+        async with httpx.AsyncClient(verify=False) as client:
```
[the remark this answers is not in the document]
JersyJ: Ah, this is deployment_service :D, not Kelvin, but I’ll still introduce a new variable for local testing.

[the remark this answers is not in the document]
JersyJ: Done, introduced DEBUG.

deploy/nginx.conf
```diff
@@ -18,7 +22,7 @@ http {
         ssl_certificate_key /etc/nginx/certs/privkey.pem;
 
         # Deny illegal Host headers
-        if ($host !~* ^(kelvin.cs.vsb.cz)$ ) {
+        if ($host !~* ^(kelvin\.cs\.vsb\.cz|nginx)$) {
```
Kobzol: The `nginx` part is here because of a health-check? Didn't we discuss that the health check would go through the production `kelvin.cs.vsb.cz` domain, so that we can test end-to-end that the app works? 🤔

JersyJ: We discussed both options, and I preferred using Nginx since it makes more sense to me. However, with this setup, you can actually choose based on the data sent in the request (healthcheck_url).

PR #683, opened 2025-05-27
Student transfer rewrite to Vue

geordi:


geordi:
I added some more types. Moving of files on the FS works for me locally.

web/urls.py
```diff
@@ -135,6 +135,7 @@
     ),
     path("tasks", teacher_view.task_list, name="tasks"),
     path("students", teacher_view.student_list, name="students"),
+    path("student/transfer", teacher_view.student_transfer, name="student_transfer"),
```
Kobzol, line 138: Is there some way to get to this link from the UI, or is it a "magic link"? :)

frontend/src/Teacher/StudentTransfer.vue
```diff
@@ -0,0 +1,258 @@
...
+
+  result.value = res.transfers;
+  busy.value = false;
+}
+</script>
+
+<template>
+  <div v-if="data">
+    <select v-model="semester">
+      <option v-for="item in Object.keys(data.semesters)" :key="item" :value="item">
```
Kobzol, line 163: image

The sorting isn't great here :) The keys should be sorted everywhere, they might be in arbitrary order.

frontend/src/Teacher/StudentTransfer.vue
```diff
@@ -0,0 +1,258 @@
...
+      class="btn"
+      :class="canTransfer ? 'btn-success' : 'btn-danger'"
+      :disabled="!canTransfer"
+      @click="transfer"
+    >
+      {{ busy ? 'Transferring...' : 'Transfer' }}
+      <span v-if="!canTransfer"
+        >(
+        <span v-if="assignments.length">
+          <span v-if="notAssignedCount > 0">{{ notAssignedCount }} tasks need to be assigned</span>
```
Kobzol, line 250: image

It might be better to put this warning somewhere on the page, as now it only appears once you click on "Transfer" :)

Kobzol (commented):
The UX looks great! The transfer (of some students from the live DB) fails for me since I don't have all the files locally. I think that the transfer should ideally be resilient against such failures, because sometimes they can happen even on production. The failure [here](https://github.com/mrlvsb/kelvin/blob/a6a1a7e2e20ae5462f5d49c2c5ab8ee2ef111493/api/views/default.py#L847) ideally shouldn't stop the transfer, IMO.

web/urls.py
geordi: Well, it's still magic but most of the users know it. Anyway, there's a big trouble in access rights so those without superuser rights basically can't transfer students.

frontend/src/Teacher/StudentTransfer.vue
geordi: Will look into that.

geordi:
OK. I updated the error handling of files. The error is now propagated into the UI. The transfer continues then.

> The UX looks great! The transfer (of some students from the live DB) fails for me since I don't have all the files locally. I think that the transfer should ideally be resilient against such failures, because sometimes they can happen even on production. The failure [here](https://github.com/mrlvsb/kelvin/blob/a6a1a7e2e20ae5462f5d49c2c5ab8ee2ef111493/api/views/default.py#L847) ideally shouldn't stop the transfer, IMO.

frontend/src/Teacher/StudentTransfer.vue
geordi: You're right. Updated in the latest forced push.

geordi:
Well, no I think I addressed all your points. I still see many ways to improve the overall UX. Thou not sure if I want to implement them since that would take another 80% of time and this functionality is used by small subset of our users.

PR #686, opened 2025-05-28
Clearly display if IP addresses if from our labs

geordi:
It's a bit stringy, bulky, and in frontend but at this point it's good enough?


Kobzol (commented):
I guess it would make a bit more sense to do it on the server and send that information to the backend, but feel free to merge as is.

geordi:
Well, now, it's good enough :-)

PR #691, opened 2025-06-03
Add Pyrefly CI tests

geordi:
This is a testing run that does not affect the current CI pipeline conclusion but we'll see outputs of Pyrefly in PRs.


.github/workflows/ci.yml
```diff
@@ -27,6 +27,25 @@ jobs:
...
+        with:
+          python-version: "3.12"
+      - name: Install dependencies
+        run: |
+          sudo apt-get install libgraphviz-dev build-essential
+          python -m pip install --upgrade pip
+          pip install uv==0.7.8
+          uv sync --frozen
+      - name: Check migrations
+        run: uv run manage.py makemigrations api common web survey quiz --check
```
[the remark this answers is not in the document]
geordi: I wasn't sure if integration into `test-backend` would make the CI not passable. So I created separate job, but in the end of the day it should be part of the `test-backend`.

geordi:
I didn't know about `ty` until now 😁 Anyway, I think we can integrate all the available tools.

> I thought we would dogfood https://github.com/astral-sh/ty 😁 But that's probably too early, so let's go with this.

.github/workflows/ci.yml
```diff
@@ -27,6 +27,9 @@ jobs:
         run: uv run ruff format --check
       - name: Check lockfile
         run: uv lock --locked
+      - name: Pyrefly Python code
+        run: uv run pyrefly check
+        continue-on-error: true
```
Kobzol, line 32: Is the GitHub annotation output the default? I noticed it was here before, but you remove it then.

geordi: GitHub format is not supported by Pyrefly, only text and JSON is.

Issue #693, opened 2025-06-18
Add support or dissable possibility to oploat `7z` files


geordi:
Sometimes, students upload a `7z` file. It would be nice to extract it like we do with `zip` files or disable possibility to upload such files.

PR #694, opened 2025-06-19
support for uploading 7z files

Spiritedswordsman:
#693 
Add 7z archive support to uploader

- Added support for extracting .7z files using py7zr

- Unified ZIP and 7z handling in a single ArchiveUploader class

- Updated pyproject.toml to include py7zr as a dependency

Let me know if you want to tag this as a fix, feature, or refactor, this is my one of the first open source contributions!

### DIFF
--- a/common/upload.py
+++ b/common/upload.py
@@ -1,6 +1,7 @@
 import os
 import re
 import zipfile
+import py7zr
 from os.path import basename, dirname
 from typing import List
 
@@ -37,20 +38,36 @@ class Uploader:
             raise Exception(f"Invalid file type {type(file)}")
 
 
-class ZipUploader(Uploader):
-    def __init__(self, file):
+class ArchiveUploader(Uploader):
+    def __init__(self, file, format):
         super().__init__()
-        self.archive = zipfile.ZipFile(file, "r")
+        self.format = format
+        self.archive = None
+
+        if format == "zip":
+            self.archive = zipfile.ZipFile(file,"r")
+        elif format == "7z":
+            self.archive = py7zr.sevenzipfile(file, mode="r")
+
 
     def get_files(self):
-        return [(f.filename, f) for f in self.archive.filelist if not f.is_dir()]
+        if self.format == "zip":
+            return [(f.filename, f) for f in self.archive.filelist if not f.is_dir()]
+        elif self.format == "7z":
+            return [(name, None) for name in self.archive.getnames()]
+        return []
 
     def upload_file(self, path: str, file, submit: Submit):
-        self.check_file_type(file, zipfile.ZipInfo)
-
         target_path = submit.source_path(path)
         os.makedirs(dirname(target_path), exist_ok=True)
-        self.archive.extract(path, path=submit.dir())
+
+        if self.format == "zip":
+            self.archive.extract(path, path=submit.dir())
+        elif self.format == "7z":
+            extracted = self.archive.read([path])
+            with open(target_path, 'wb') as f:
+               f.write(extracted[path].read())
+
         self.count += 1
         return target_path
 
@@ -120,8 +137,10 @@ class TooManyFilesError(BaseException):
 
 
 def upload_submit_files(submit: Submit, paths: List[str], files: List[UploadedFile]):
-    if len(paths) == 1 and get_extension(paths[0]) == ".zip":
-        uploader = ZipUploader(files[0])
+    if len(paths) == 1 and (zipfile.is_zipfile(files[0]) or py7zr.is_7zfile(files[0])):
+        format = ("zip","7z")[py7zr.is_7zfile(files[0])]
+        uploader = ArchiveUploader(files[0],format)
+
     else:
         uploader = FileUploader(paths, files)
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,6 +42,7 @@ dependencies = [
     "Unidecode==1.3.6",
     "uWSGI==2.0.24",
     "django-ipware==7.0.1",
+    "py7zr==1.0.0"
 ]
 
 [tool.uv]


common/upload.py
```diff
@@ -37,20 +38,36 @@ def check_file_type(self, file, type):
             raise Exception(f"Invalid file type {type(file)}")
 
 
-class ZipUploader(Uploader):
-    def __init__(self, file):
+class ArchiveUploader(Uploader):
```
Kobzol, line 41: Instead of merging two functionalities into one class, could you just create a separate class, e.g. `SevenZipUploder`?

common/upload.py
```diff
@@ -120,8 +137,10 @@ class TooManyFilesError(BaseException):
 
 
 def upload_submit_files(submit: Submit, paths: List[str], files: List[UploadedFile]):
-    if len(paths) == 1 and get_extension(paths[0]) == ".zip":
-        uploader = ZipUploader(files[0])
+    if len(paths) == 1 and (zipfile.is_zipfile(files[0]) or py7zr.is_7zfile(files[0])):
+        format = ("zip","7z")[py7zr.is_7zfile(files[0])]
```
Kobzol, line 141: This is quite hard to read :) Just use `if`, please.

Kobzol (commented):
Thank you! Left some comments. You should also commit the updated `uv.lock` file.

common/upload.py
Spiritedswordsman: sure, I thought that too, but this made more sense at that time... but seems it will be cleaner if I separate them out

common/upload.py
```diff
@@ -40,23 +41,44 @@ def check_file_type(self, file, type):
 class ZipUploader(Uploader):
     def __init__(self, file):
         super().__init__()
-        self.archive = zipfile.ZipFile(file, "r")
+        self.archive = self.archive = zipfile.ZipFile(file,"r")
```
Kobzol, line 44: Duplicated assignment.
```suggestion
        self.archive = zipfile.ZipFile(file, "r")
```

common/upload.py
```diff
@@ -40,23 +41,44 @@ def check_file_type(self, file, type):
...
 
     def get_files(self):
         return [(f.filename, f) for f in self.archive.filelist if not f.is_dir()]
 
     def upload_file(self, path: str, file, submit: Submit):
-        self.check_file_type(file, zipfile.ZipInfo)
```
Kobzol, line 49: Why remove this?

common/upload.py
```diff
@@ -40,23 +41,44 @@ def check_file_type(self, file, type):
...
         self.count += 1
         return target_path
 
     def close(self):
         self.archive.close()
 
+class SevenZUploader(Uploader):
+    def __init__(self, file):
+        super().__init__()
+        self.archive = self.archive = py7zr.sevenzipfile(file, mode="r")
```
Kobzol, line 63: Same here.

common/upload.py
Spiritedswordsman: the reason was file type was begin checked before choosing the uploader class, twice already, seemed 3 times would be a overkill. (though I'll change it as you say, clearly your senior and more experienced)
also In zipfile, files in the archive are represented by zipfile.ZipInfo objects.
In py7zr, there's no direct equivalent class like ZipInfo. The getnames() and getinfo() methods return basic types like str or dictionaries.

Kobzol: Before we only checked the extension, so this wasn't redundant, if something has `.zip` extension, that doesn't mean it's a ZIP file :)

`self.check_file_type(file, str)` in the 7zip file definitely won't work, `str` is a built-in function. Let's remove that.

Kobzol:
Also please reformat code with `ruff format`, thanks!

common/upload.py
```diff
@@ -120,8 +145,11 @@ class TooManyFilesError(BaseException):
 
 
 def upload_submit_files(submit: Submit, paths: List[str], files: List[UploadedFile]):
-    if len(paths) == 1 and get_extension(paths[0]) == ".zip":
-        uploader = ZipUploader(files[0])
+    if len(paths) == 1 and (zipfile.is_zipfile(files[0]) or py7zr.is_7zfile(files[0])):
```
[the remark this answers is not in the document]
Spiritedswordsman: Heyy, made new changes, let me know how it goes!

common/upload.py
```diff
@@ -37,6 +39,15 @@ def check_file_type(self, file, type):
         if not isinstance(file, type):
             raise Exception(f"Invalid file type {type(file)}")
 
+    def check_7z_file(self, file):
+        if isinstance(file, UploadedFile):
```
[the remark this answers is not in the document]
Spiritedswordsman: okay then

common/upload.py
```diff
@@ -58,6 +61,29 @@ def close(self):
         self.archive.close()
 
 
+class SevenZUploader(Uploader):
+    def __init__(self, file):
+        super().__init__()
+        self.archive = py7zr.sevenzipfile(file, mode="r")
```
Kobzol, line 67: module 'py7zr' has no attribute 'sevenzipfile'

.. Have you tried this locally? Or are you copy-pasting AI output here? :)

Kobzol (commented):
Please try if it works locally before sending changes.

common/upload.py
Kobzol: Did you build the frontend? https://github.com/mrlvsb/kelvin/blob/master/docs/setup.md#building-the-frontend Try running `npm run dev` in the `frontend` directory.

Spiritedswordsman: image

that doesn't change much apart from the frontend, I just want to reach the part where I could upload the 7z type file and test it

Kobzol: Yeah, we sadly don't yet have a good story for priming the database. You'd have to create the classes manually in the `/admin` section.

Issue #689, opened 2025-06-02
Mark submit as final


geordi:
Have an option to mark final submission (no further submissions allowed, or at least is seen that such submission is sent after this marked submission), also, such flag can be set by teacher.

Teacher has to approve this request.

Issue #586, opened 2025-01-30
Allow submission uploads from certain IP adress range


geordi:
When writing tests in classrooms, allow submit only from predefined IP address range. This also requires to be able to manually add IP address by the teacher.

Issue #408, opened 2023-12-18
Add suport for "Test" task


geordi:
This task would be useful for test during or at the end of the semester. It would be nice to have features like:

- request by student to display assignment and this needs to be approved by teacher
- have option to mask final submission (no further submissions allowed, or at least is seen that such submission is sent after this marked submission), also, such flag can be set by teacher

PR #705, opened 2025-09-15
L&T - test task, mark submit as final, allow certain IP address range only

SimonAdamek:
- new "test" task type
- last submit can be marked as final; after that no more submits are allowed and the solution is hidden from student
- in teacher overview, tests with submits but without a final mark are shown as '!'
- support setting an IP address range for tests; access from other address returns 403 page
- mark of a submit is displayed in event log

Related to #408 
Related to #586
Related to #689

### DIFF
--- a/api/views/default.py
+++ b/api/views/default.py
@@ -34,6 +34,7 @@ from common.event_log import (
     UserEventSubmit,
     UserEvent,
     UserEventTaskDisplayed,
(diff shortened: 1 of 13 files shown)


geordi (changes requested):
As I already wrote in a comment... Please split your work into more PRs. We'll iterate on that.

PR #707, opened 2025-09-16
Update Cargo Docker image to 1.90.0

Kobzol:
Shouldn't be merged until the 1.90.0 Docker image is actually released (should be Thursday/Friday this week).


geordi:
1.90.0 Docker image is ready, so merging and rebuilding images on servers.

PR #708, opened 2025-09-20
Fix of edit link in Class Detail

### LINKED ISSUE #706: Broken task edit link
After loading the initial page of teacher's Kelvin, everything is fine. Using a task type filter in class header renders edit link in hoover box unusable. It somehow removes `#` character from the link. Please @Dym03 can you look at it since you're the author ;-) Thanks!

Dym03:
I wasn't able to recreate the problem on the local, for me it worked without the hash, but this should fix the problem.
Seems like the use:link have problem with rerendering in use with the spa-router, please correct me if I am wrong :). 

Fixes #706

### DIFF
--- a/frontend/src/ClassDetail.svelte
+++ b/frontend/src/ClassDetail.svelte
@@ -222,7 +222,7 @@ let showSummary = false;
                       </a>
                       <div class="more-content border shadow rounded bg-body p-1">
                         {assignment.name}
-                        <a href="/task/edit/{assignment.task_id}" use:link title="Edit"
+                        <a use:link={`/task/edit/${assignment.task_id}`} title="Edit"
                           ><span class="iconify" data-icon="clarity:edit-solid"></span></a>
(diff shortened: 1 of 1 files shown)


Kobzol:
Wtf, it really only happens for "Laboratory", and not for "Homework" :laughing: @Dym03 Any idea what could be causing that difference?

Dym03:
For me it does that for homework also, seems like it is not consistent because sometimes it doesn't break the link. I was informed about the broken link. What I meant with unable to recreate the issue, is that if i click on the "broken link" it redirects me to the correct page as if the link wasn't broken. 

My idea why this might happen is the use:link as i mentioned earlier, but I am not used to this technology as much so i can't say I am sure. The solution in this PR should solve this issue then even at the cost of the fast switch you mentioned earlier.

Dym03:
Sure will do :), thanks for the deep dive and explanation.

PR #711, opened 2025-09-25
Mark submit as final button

SimonAdamek:
Adds a button to mark exam task as final. Marked tasks have hidden source code and disabled upload.  Marking is also recorded in event log.

Related to #689

### DIFF
--- a/common/submit.py
+++ b/common/submit.py
@@ -50,34 +50,33 @@ def store_submit(request: HttpRequest, assignment: AssignedTask) -> Submit:
     ):
         raise SubmitPastHardDeadline("Submit was submitted after deadline")
 
-    # get existing submits of the student
-    submits = Submit.objects.filter(assignment__pk=assignment.id, student__pk=request.user.id)
+    # Get existing submits of the student
+    submits = list(
+        Submit.objects.filter(assignment__pk=assignment.id, student__pk=request.user.id).order_by(
+            "-created_at"
+        )
+    )
 
     # Check submit date across all tasks, just to be a bit more defensive
-    last_submit = Submit.objects.filter(student=request.user).order_by("-created_at").first()
+    last_student_submit = (
+        Submit.objects.filter(student=request.user).order_by("-created_at").first()
+    )
 
-    if last_submit is not None:
-        since_last_submit = datetime.datetime.now(datetime.timezone.utc) - last_submit.created_at
+    if last_student_submit is not None:
+        since_last_submit = (
+            datetime.datetime.now(datetime.timezone.utc) - last_student_submit.created_at
+        )
         if since_last_submit < SUBMIT_RATELIMIT:
             raise SubmitRateLimited("Submit was rate limited", SUBMIT_RATELIMIT - since_last_submit)
 
-    last_submit = (
-        Submit.objects.filter(assignment__pk=assignment.pk, student=request.user)
-        .order_by("-created_at")
-        .first()
-    )
-
-    if last_submit is not None and last_submit.is_final:
+    if any(s.is_final for s in submits):
         raise SubmitAfterFinal("Attempt to create submit after marking previous as final")
 
     # TODO: transaction/better submit_num checking
     s = Submit()
     s.student = request.user
     s.assignment = assignment
-    s.submit_num = (
-        Submit.objects.filter(assignment__id=s.assignment.id, student__id=request.user.id).count()
-        + 1
-    )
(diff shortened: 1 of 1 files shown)


templates/web/task_detail.html
```diff
@@ -119,6 +119,8 @@ <h5 title="Top 5 plagiarism matches for some checked submit of this student" cla
             </li>
             {% endif %}
             {% if submit %}
+
+            {% if is_teacher or task.type != 'exam' or not submits.first.is_final %}
```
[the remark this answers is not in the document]
Kobzol: A bit confusing that they are ordered in the reverse way, but let's keep it. Thanks.

api/views/default.py
```diff
@@ -473,7 +484,7 @@ def event_list(request: HttpRequest, login: str):
             created_at=event.created_at,
         )
         result.append(data)
-
+    print(result)
```
Kobzol: Forgotten print (just to make sure we don't leave it there).

common/submit.py
```diff
@@ -49,11 +54,13 @@ def store_submit(request: HttpRequest, assignment: AssignedTask) -> Submit:
...
     # Check submit date across all tasks, just to be a bit more defensive
-    last_submit_date = Submit.objects.filter(student=request.user).order_by("-created_at").first()
-    if last_submit_date is not None:
-        since_last_submit = (
-            datetime.datetime.now(datetime.timezone.utc) - last_submit_date.created_at
-        )
+    last_submit = Submit.objects.filter(student=request.user).order_by("-created_at").first()
+
+    if last_submit is not None:
+        if assignment.task.type == "exam" and last_submit.is_final:
```
Kobzol: Let's check only `is_final` and not the task type, to avoid scattering the logic for what does a final submit mean across the codebase. We will make sure elsewhere that only exam submits can be marked as final (for now).

web/views/student.py
```diff
@@ -1078,6 +1073,34 @@ def upload_results(request, assignment_id, submit_num, login):
     return JsonResponse({"success": True})
 
 
+def mark_solution_as_final(request, assignment_id, login, submit_num):
+    submit = get_object_or_404(
```
Kobzol: This allows anyone to mark any submit as being final. Let's:
- Add `@login_required`, to ensure that only logged in users can do that
- Check that the owner of the submit (`login`) matches the username of the logged in user, otherwise raise `PermissionDenied` (see e.g. `task_detail`) for an example.
- Check that the submit being marked as final is the most recent one for the given assignment and student.

SimonAdamek:
OK, I removed the source code hiding in the final tasks feature. I have some notes from one of the consultations about it — I guess I just misunderstood it.

web/views/student.py
```diff
@@ -1078,6 +1073,48 @@ def upload_results(request, assignment_id, submit_num, login):
...
+
+    last_submit = (
+        Submit.objects.filter(
+            assignment__pk=assignment_id,
+        )
+        .order_by("-created_at")
+        .first()
+    )
+
+    if submit.pk == last_submit.pk:
```
Kobzol: Let's raise a 400 HTTP exception if the condition doesn't hold.

templates/web/task_detail.html
```diff
@@ -151,11 +151,11 @@ <h5 title="Top 5 plagiarism matches for some checked submit of this student" cla
 
                     {% if task.type == 'exam' and submit != null and submit.submit_num == submits.first.submit_num %}
                       {% if submits.first.is_final %}
-                          <div class="me-1" title="Link to the next student thas has unevaluated submit(s) for the same task.">
+                          <div class="me-1" title="Allow student to create new submits. ">
                             <a class="btn btn-sm btn-info" href="{% url 'unmark_solution_final_mark' submit.assignment.pk submit.student submit.submit_num%}">Unmark final flag</a>
                           </div>
                       {% else %}
-                          <div class="me-1" title="Link to the next student thas has unevaluated submit(s) for the same task.">
+                          <div class="me-1" title="Prevents the student from sending further submits.">
```
Kobzol: This title is shown to the student, not the teacher, so let's tell them what does it mean:
"Once you mark a submit as final, you won't be able to create any additional submit for this task.".

SimonAdamek: I'm pretty sure that the title is displayed to the teacher because it's behind the "Next student to evaluate" button. The text you mentioned ("Once you mark a submit as final...") is displayed to the student in a modal window when they try to mark the submission as final.

PR #713, opened 2025-09-27
Make the "Next student to evaluate" link point directly to source code

Kobzol:
There should always be a source code available, because we know that there is at least a single submit for the given student, otherwise the link wouldn't point to it.

Suggested by Pavel Dohnálek.


geordi:
Nice, I wanted to point it out as well :-)

PR #724, opened 2025-10-23
Ip filtering

SimonAdamek:
Allow submission uploads and task viewing only from defined Classrooms IP ranges.

Models:

- new model ClassroomIpRange - stores ip range of one classroom

EditTask page:

- Added a new field displaying a list of classrooms (only for exam task type).
- The list is searchable and sorted by checked/unchecked status, then alphabetically.

IP-based access control:

- Added decorator to web/views/student.py 
- The API also checks IP addresses to prevent direct POST requests from unauthorized clients.

Administrator page:

- Added custom form to add new classrooms based on IP address ranges or CIDR notation.

Related to #586


Kobzol:
Hi, thanks for the PR! This looks really cool.
 
There is some functionality that is not strictly related to IP filtering, so it would be great to merge it separately. Could you split the change with the custom error pages to a separate PR please? In the error page it would be nice to use the normal layout and make it a "normal page", with the header, notifications and everything, so that it is not just a bare white page.

common/utils.py
```diff
@@ -81,3 +86,39 @@ def get_client_ip_address(request: HttpRequest) -> IPAddressString | None:
         return None
     else:
         return IPAddressString(client_ip)
+
+
+def ip_address_check(function):
```
Kobzol: Could you please add a doc comment here that explains what this decorator does? It's pretty magical :) In particular, we should document how it interacts with a parameter named `assignment_id`.

PR #727, opened 2025-10-28
First LLM implementation

Firestone82:
This is first working implementation of kelvin llm implementation.
Currently its for testing purposes visible only to teachers with static view.

### DIFF
--- a/common/evaluate.py
+++ b/common/evaluate.py
@@ -12,7 +12,7 @@ from django.core import signing
 from django.urls import reverse
 from django.utils import timezone
 
-from common.summary.summary import summarize_submit
+from common.ai_summary.summary import summarize_submit
 from common.utils import is_teacher
 from evaluator.evaluator import Evaluation
 from evaluator.testsets import TestSet
--- a/frontend/src/TaskDetail.svelte
+++ b/frontend/src/TaskDetail.svelte
@@ -224,7 +224,7 @@ function countComments(comments) {
   };
   for (const line of Object.values(comments)) {
     for (const comment of Object.values(line)) {
-      if (comment.type === 'automated' || comment.type === 'summary') {
+      if (comment.type === 'automated' || comment.type === 'ai-review') {
         counts.automated += 1;
       } else {
         counts.user += 1;
--- a/web/views/llm.py
+++ b/web/views/llm.py
@@ -5,7 +5,7 @@ from django.shortcuts import get_object_or_404
 from django.views.decorators.csrf import csrf_exempt
 
 from common.models import Submit
-from common.summary.summary import SUMMARY_RESULT_FILE_NAME
+from common.ai_summary.summary import AI_REVIEW_RESULT_FILE_NAME
 from web.views.utils import authenticate_submit_token_request
 
 
@@ -26,7 +26,7 @@ def post_submit_summary_result(request, assignment_id, login, submit_num):
     )
 
     os.makedirs(result_path, exist_ok=True)
-    summary_file_path = os.path.join(result_path, SUMMARY_RESULT_FILE_NAME)
+    summary_file_path = os.path.join(result_path, AI_REVIEW_RESULT_FILE_NAME)
 
     with open(summary_file_path, "wb") as summary_file:
         summary_file.write(request.body)
--- a/web/views/student.py
+++ b/web/views/student.py
@@ -39,6 +39,12 @@ from notifications.models import Notification
 from notifications.signals import notify
 from serde.json import from_json
 
+from common.ai_summary.dto import AIReviewResult
+from common.ai_summary.summary import (
+    AI_REVIEW_RESULT_FILE_NAME,
+    AI_REVIEW_COMMENT_TYPE,
+    AI_REVIEW_COMMENT_AUTHOR,
+)
 from common.evaluate import get_meta
 from common.event_log import record_task_displayed, record_final_submit_event
 from common.models import (
@@ -52,8 +58,6 @@ from common.models import (
 )
 from common.plagcheck.moss import PlagiarismMatch, moss_result
 from common.submit import SubmitRateLimited, store_submit, SubmitPastHardDeadline
-from common.summary.dto import ReviewResult
-from common.summary.summary import SUMMARY_RESULT_FILE_NAME
 from common.upload import MAX_UPLOAD_FILECOUNT, TooManyFilesError
 from common.utils import is_teacher
 from evaluator.results import EvaluationResult
@@ -226,7 +230,7 @@ def student_index(request):
 
 def get_submit_data(submit: Submit) -> SubmitData:
     results = []
-    summary = ReviewResult("", [])
+    ai_review = AIReviewResult("", [])
 
     try:
         results = EvaluationResult(submit.pipeline_path())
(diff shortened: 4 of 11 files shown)


web/views/llm.py
```diff
@@ -0,0 +1,33 @@
...
+@csrf_exempt
+def post_submit_summary_result(request, assignment_id, login, submit_num):
+    if request.method != "POST":
+        return JsonResponse({"error": "Invalid request method."}, status=405)
+
+    submit = get_object_or_404(
+        Submit, assignment_id=assignment_id, student__username=login, submit_num=submit_num
+    )
+
+    has_permission_for_submit(request, submit)
```
Kobzol: Let's use the same authentication mechanism here as we do for `upload_results` (feel free to move it into a separate method, e.g. `authenticate_evaluator_request` or something). We should not allow anyone else to authenticate here.

web/views/student.py
```diff
@@ -223,16 +225,29 @@ def student_index(request):
...
         pass
 
-    data = {
-        "submit": submit,
-        "results": results,
-    }
+    try:
+        with open(os.path.join(submit.pipeline_path(), "summary.json")) as f:
+            result_dict = json.load(f)
+            summary = dict_to_dataclass(result_dict, ReviewResult)
```
Kobzol: As noted earlier, this is not needed, please use `@serde` for the LLM summary models.

web/views/student.py
```diff
@@ -808,6 +823,40 @@ def dump_comment(comment):
         except KeyError as e:
             logging.exception(e)
 
+    # add comments from llm summary
+    if is_teacher(request.user):  # Currently only teachers can view LLM summary comments
+        llm_summary = getattr(resultset["summary"], "summary", "")
```
Kobzol: Could you please refactor the `get` function in a separate PR first so that it returns a properly typed dataclass? This is becoming too brittle..

frontend/src/TaskDetail.svelte
```diff
@@ -335,6 +343,10 @@ window.addEventListener('hashchange', goToSelectedLines);
             <span class="badge bg-primary" title="Automation comments" style="font-size: 60%;"
               >{comments.automated}</span>
           {/if}
+          {#if comments.summary > 0}
```
Kobzol: We have a button for switching between automated/user/all comments:
image

It should be updated to also handle summary comments.

Although, I'm not really sure if we need to distinguish automated and summary comments.. I would treat both as "automated", to avoid introducing yet another comment category, it's already quite complicated.

common/summary/models.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+class Severity(Enum):
+    CRITICAL = "critical"
+    HIGH = "high"
+    MEDIUM = "medium"
+    LOW = "low"
+
+
+@dataclass
```
Kobzol: This should all use `@serde`.

common/summary/summary.py
```diff
@@ -0,0 +1,118 @@
...
+    return embedded
+
+
+@django_rq.job
+def summary_job(submit_url, token) -> None:
+    logging.basicConfig(level=logging.DEBUG)
+    logging.info(f"Summarizing {submit_url}")
+
+    with tempfile.TemporaryDirectory() as workdir:
+        os.chdir(workdir)
```
Kobzol: Why is this needed? I know that Kelvin does this at a few places, but it is a horrible hack that already led to annoying bugs. It shouldn't be needed to change the whole proecss working directory just to extract an archive. (not to mention that the code currently doesn't even reset the working directory back).

common/summary/summary.py
```diff
@@ -0,0 +1,118 @@
...
+
+    # Loop through downloaded files
+    for root, _, files in os.walk(source_files_path):
+        for file in files:
+            filepath = os.path.join(root, file)
+            relpath = os.path.relpath(filepath, source_files_path)
+
+            language = detect_language(file)
+            if not language:
+                logging.warning(f"Skipping file with undetected language: {relpath}")
```
Kobzol: It's not being skipped currently, missing `continue`?

common/summary/summary.py
Firestone82: The reason its there is the only one. That i copied it from different place without really thinking about it.

common/evaluate.py
```diff
@@ -1,19 +1,41 @@
...
-from common.utils import is_teacher
 
+from common.summary.summary import summarize_submit
+from common.utils import is_teacher
+from evaluator.evaluator import Evaluation
 from evaluator.testsets import TestSet
+from kelvin.settings import BASE_DIR
+
+
+def load_task_config(task_path) -> Optional[dict]:
```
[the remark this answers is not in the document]
Kobzol: I think tha calling code just passes a `str`, so let's use that. We should be using `Path`s, but we don't do that on a million other places in Kelvin.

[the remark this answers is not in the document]
Firestone82: Yes, I think that I tried that. And it still was throwing warns that I’m changing one type to another. But will check that later.

frontend/src/TaskDetail.svelte
```diff
@@ -198,6 +198,11 @@ async function load() {
   files = json['sources'].map((source) => new SourceFile(source));
   summaryComments = json['summary_comments'];
 
+  // Append summary to summary comments
```
Kobzol: Looks like this is outdated? The summary comments should already be in `summary_comments`.

Kobzol (commented):
I realized that currently we output comments as raw HTML, with doing HTML sanitization using DOMPurify. I don't feel comfortable outputting LLM output to the page through this process. Could you please add some parameter to the Vue comments that will print them "raw", i.e. using `{ ... }` and not `{@html ... }`? Also to make sure that the output won't contain any hidden HTML comments or something like that. Thanks!

web/views/utils.py
```diff
@@ -15,10 +15,21 @@ def file_response(file, filename: str, mimetype: str) -> HttpResponse:
...
     if "token" in request.GET:
         token = signing.loads(request.GET["token"], max_age=3600)
-        if token.get("submit_id") != submit.id:
+        if token.get("submit_id") == submit.id:
+            return None
+
+    raise PermissionDenied()
+
+
+def authenticate_submit_user_request(request, submit: Submit, allow_token=False) -> None:
```
Kobzol: This doesn't seem to be used anywhere.

Firestone82:
> Please rebase this on top of #731 once it gets merged.
> 
> ```
> git fetch origin
> git rebase origin/master
> ```

Is it required? From what I see in commits, it should be fine. Since I did it yesterday and there are no more commits?

common/summary/summarizer.py
```diff
@@ -0,0 +1,150 @@
...
+    Remove comments from source code based on the programming language.
+    Leaves the code structure intact. Each comment is replaced with empty lines to preserve line numbers.
+    """
+
+    if language in ["c", "cpp", "java", "javascript", "typescript"]:
+        content = re.sub(r"//.*", "", content)
+        content = re.sub(r"/\*.*?\*/", "", content, flags=re.DOTALL)
+    elif language == "python":
+        content = re.sub(r"#.*", "", content)
+        content = re.sub(r'"""(.*?)"""', "", content, flags=re.DOTALL)
```
Kobzol: This is not really a general solution. `"""` are normal multiline strings in Python, this would just remove them 😆

Have you tried telling the LLM to ignore comments in code?

Kobzol: Ok, then please remove the comments deletion for now.

web/views/student.py
```diff
@@ -223,13 +226,25 @@ def student_index(request):
 
 def get_submit_data(submit: Submit) -> SubmitData:
     results = []
+    summary = ReviewResult("", [])
```
Kobzol:
```suggestion
    ai_review = ReviewResult("", [])
```

PR #731, opened 2025-10-30
Changed submit get function, to return Dataclass instead of Dict.

Firestone82:
Requested in PR #727

### DIFF
--- a/web/views/student.py
+++ b/web/views/student.py
@@ -1,4 +1,3 @@
-import dataclasses
 import hashlib
 import io
 import json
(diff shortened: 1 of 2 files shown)


web/views/student.py
```diff
@@ -407,7 +404,7 @@ def task_detail(request, assignment_id, submit_num=None, login=None):
         )
 
     if current_submit:
-        data = {**data, **get(current_submit)}
+        data = {**data, **get(current_submit).__dict__}
```
Kobzol: :D
```suggestion
        data = {**data, "submit": current_submit, "results":  get_submit_data(current_submit).results}
```

web/models.py
geordi: I'd rather rename this file to `dto.py` so it does not interfere with *Django models* naming scheme.

Firestone82: Alright.

web/views/student.py
Kobzol: Oh, sure, having the `data` object be a dataclass would indeed be nice. That's for a separate PR though, and it's a much larger refactoring :)

PR #733, opened 2025-10-31
Add designed error pages

SimonAdamek:
Adds templates for error pages 400, 403, and 404 (i.e., BadRequest, PermissionDenied, and Http404 exceptions).

Is this style okay, or should I add a header (kelvin title and notifications)?

Screenshot 2025-10-31 at 09-32-10 Kelvin 400


web/views/common.py
```diff
@@ -10,6 +10,39 @@
...
+    }
+    return render(request, "error_page.html", context)
+
+
+def custom_404_page(request, exception=None):
+    previous_url = request.META.get("HTTP_REFERER", "/")  # return to last url or home
+    context = {
+        "status_code": 404,
+        "title": "Page Not Found",
+        "message": str(exception) if exception else "The page you requested does not exist.",
```
Kobzol: Using `str` here produces this:

Image

(The styling is also off for some reason, as the CSS files are loaded with HTML mimetype, not sure why).

Kobzol:
We also need to ensure that the HTML version of the error won't be shown when doing API requests programatically (the `api` module). Just noting this here so that I don't forget.

web/views/common.py
SimonAdamek: Screenshot 2025-10-31 at 10-39-29 Kelvin 404

SimonAdamek:
Here is the first attempt.

 Currently, there is a small issue; I have not been able to find a solution to separate Django error messages (page not found for localhost:8000/foo) from those generated by the user (raise Http404 call). So I created my own set of exceptions. However, when calling "get_object_or_404," the general text "The page you requested does not exist" is returned, instead of "No 'model' matches the query", but this probably doesn't matter outside of debugging.

This also means that it is necessary to rewrite the use of Http404/PermissionDenied/BadRequest exceptions to call a custom ones. I did this for the student.py file. I don't know if it needs to be changed elsewhere.

SimonAdamek:
Oh, sorry, I didn't make that clear enough.

It is still possible to use Django exceptions such as Http404, and a new page will appear. The only difference is: if the Django exception is used, the text is replaced with a generic text (such as "Not found"). If you want your own text in error page, you have to use my exception.

The reason is, that I can't tell whether the Http404 exception was thrown by the user or the system, and the system exception may contain unnecessarily long text (as you showed above).

I rewrote all the exceptions  in student.py just so that the two types wouldn't get mixed up.

Kobzol:
When trying to load CSS/JS files (e.g. static/frontend.3835799c3319.js), the Django server returns 200 for them, but the returned content is the 404 error HTML page. I'll try to take a look at why that happens.

Kobzol:
I spent several hours on this piece of crap. Seriously, whoever designed this part of Django had to be drunk. This is such a stupid mechanism that doesn't allow handling errors in an unified way..

I agree that we should just raise custom exceptions. I don't like it, but it will be the easiest to handle in Django. I'm seriously starting to think that Django is not a serious web framework... What a mess.

I pushed some changes that I made during my experiments and rebased due to a conflict. Let's try what this does in production.

web/views/common.py
```diff
@@ -1,23 +1,24 @@
...
-        return render(request, "error_page.html", exception)
-
-    raise exception
+def render_custom_error_page(request: HttpRequest, exception):
+    exception = HttpExceptionData.from_exception(exception)
+    ctx = dict(
+        status_code=exception.status,
+        message=exception.message,
+    )
+    return render(request, "error_page.html", ctx, status=exception.status)
```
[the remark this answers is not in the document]
Kobzol: Good point. However, I tried it before this PR and doing an API call to e.g. `/api/foo` still produced HTML content xD So it's not even a regression. The API backend really needs a rewrite, ideally using django-ninja...

PR #743, opened 2025-11-04
[Quiz] Use `pk` instead of `id` in DB models

geordi:


Kobzol (approved):
This is just stylistic, right?

Issue #603, opened 2025-02-05
Disable viewing of other tasks and submits for students that write a test


geordi:
During tests students won't be able to see other tasks and own submits. Someone else can upload it on their behalf and the student then use this upload as his own during the test.

PR #745, opened 2025-11-06
Create prohibit_during_test decorator

SimonAdamek:
Teachers sometimes allow students to upload materials for tests to Kelvin. So, I enabled viewing of assignments with a "Hard deadline" that expires before the start of the test. 

This means that if a teacher wants to allow students to upload test materials to Kelvin, they must create a task with a "hard deadline" before the test. It's a bit complicated, but I couldn't think of any other way to solve it.

Related to #603

### DIFF
--- a/common/utils.py
+++ b/common/utils.py
@@ -147,7 +147,7 @@ def prohibit_during_test(function):
         if is_teacher(request.user):
             return function(*args, **kwargs)
 
-        active_exams = get_active_exams_at(request.user, datetime.now(), timedelta(0))
+        active_exams = get_active_exams_at(request.user, datetime.now(), timedelta(0.5))
 
         if not active_exams:
             return function(*args, **kwargs)
--- a/common/utils.py
+++ b/common/utils.py
@@ -1,14 +1,16 @@
 import io
 import re
 import tarfile
-from datetime import timedelta
+from datetime import timedelta, datetime
 from functools import lru_cache
 from typing import NewType
 
 import django.contrib.auth.models
 import requests
 from django.conf import settings
+from django.core.exceptions import PermissionDenied
 from django.http import HttpRequest
+from django.http.response import Http404
 from ipware import get_client_ip
 
 from .inbus import inbus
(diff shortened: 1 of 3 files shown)


Kobzol:
I think you already talked about this with geordi, but there are some "edge cases" that we should consider, for example someone uploading a submit e.g. a week before a test, and then reading its code, or someone submitting code through a comment on a different task, and then deleting that comment to hide tracks.

We were talking about making it impossible to delete and modify comments for students during a test. However, that is not enough, because someone could make a comment during a test, then copy the output from it, and then delete the comment just after the test ends (but before someone realistically has a chance to take a look at their submit to see the incriminating comment).

Maybe what we could do is to make any comments that are created ~hour before/after a test be "pinned", so that they cannot be modified nor removed by the student, ever? @geordi

Kobzol:
The problem with "during test" is that quite often tests last longer than the deadline :)

SimonAdamek:
I've made the necessary changes based on your feedback—access is now limited to the test page only, the test has a time window of +/- 30 minutes, and comments are disabled in the new API.

I apologize for the unnecessary commits; I don't know how I managed to make this mistake

Kobzol:
Could you rebase the PR please? `git fetch origin && git rebase origin/master`.

common/utils.py
```diff
@@ -123,3 +124,50 @@ def build_evaluation_download_uri(request, location):
...
+            assignment_id = None
+        else:
+            author = args[0].user
+            assignment_id = kwargs.get("assignment_id")
+
+        # Allways allow teacher access
+        if is_teacher(author):
+            return function(*args, **kwargs)
+
+        active_exams = get_active_exams_at(author, datetime.now(), timedelta(0.5))
```
Kobzol: `timedelta(0.5)` creates a time window of half a day, not half an hour :)

Kobzol:
As a reminder, once this PR is merged, students won't be able to access any other task when they have an active exam, in the interval [30 minutes before exam, 30 minutes after exam].

@geordi Are you fine with that logic?

PR #751, opened 2025-11-12
LLM - Second wave of changes

Firestone82:
This PR ended up bigger than usual since it contains changes that were waiting on the previous PR.


Kobzol:
Let's split this up into multiple smaller PRs, this one is very large, and already has conflicts. Let's start with the validation of the `config.yml` file, that should be reasonably self-contained.

PR #752, opened 2025-11-14
Add LLM config validation to FE

Firestone82:
@Kobzol Separated PRs will follow. This is just the configuration of LLM.


frontend/src/PipelineValidation.js
```diff
@@ -293,16 +294,68 @@ class UnionRule {
     }
 }
 
+// Character-class flags
+const ValueRuleFlags = {
```
Kobzol: What is this for?

Kobzol: There is nothing using these validations in this PR, so remove it please. We can add something like this once it actually becomes needed, although I would prefer a more composable solution.

PR #753, opened 2025-11-14
LLM Front-end implementation

Firestone82:
This PR implement LLM FE. It should be written in Vue rather then Svelte, but since I have no clue how it works, I rather added it into already existing solution.

### DIFF
--- a/frontend/src/CodeRow.svelte
+++ b/frontend/src/CodeRow.svelte
@@ -3,6 +3,7 @@ import { createEventDispatcher } from 'svelte';
 import CommentForm from './CommentForm.svelte';
 import Comment from './Comment.svelte';
 import { user } from './global.js';
+import SuggestedComment from './SuggestedComment.svelte';
 
 export let line;
 export let lineNumber;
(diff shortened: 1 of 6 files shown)


geordi:
Can be the `SuggestedCommend` component written in Vue since it's a new component?

geordi:
> > Can be the `SuggestedCommend` component written in Vue since it's a new component?
> 
> I tried to, but that requires to rewrite 4 components only because of comment "clone". So I chose this instead. I would rather to not see this Svelte, Vue shenanigans at all. 😥

Well, what we're trying to achieve is to make Kelvin a better place to work in. You're working on a nice addition. Unfortunately, there are parts of  Kelvin that need to be worked on and we're somehow in a transition from Svelte to Vue. I understand your frustration but we can't really add new features (and continue in an unmaintainable state) and sacrifice a long term plan of transition.

So whenever you want to add something, please consider also rewriting affected parts into a new state (being Vue). I know, you can do it. This is exactly the situation that many developers are facing in real life. I think it's worth it. It took me personally a while to accept this attitude, but it's better now.

PR #755, opened 2025-11-15
Docosaurus initialization - better docs

Firestone82:
First thing that hit me in the face when I started working on kelvin was huge gap in documentation its clarity and localization.

So this should be a huge enhancement on how docs are handeled and how users can see this project. 

Currently there is missing sub READMES like pipeline or external_open. But this will come soon.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,82 @@
+name: Deploy Docs to GitHub Pages
+
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - "docs/**"
+      - ".github/workflows/docs.yml"
+
+  pull_request:
+    paths:
+      - "docs/**"
+      - ".github/workflows/docs.yml"
+
+  workflow_dispatch: {}
+
+jobs:
+  lint-mdx:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Setup Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20
+
+      - name: Install markdownlint
+        run: npm install -g markdownlint-cli
+
+      - name: Run Markdownlint on MDX files
+        run: |
+          cd docs
+          npx markdownlint "**/*.mdx" "**/*.md"
+
+  build:
+    runs-on: ubuntu-latest
+    needs: lint-mdx
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -12,12 +12,4 @@ VSB-TUO is deployed at `https://kelvin.cs.vsb.cz`.
...
-You can find documentation on how Kelvin works and how to contribute to it in the following files:
-- [Architecture](docs/architecture.md) describes the high-level architecture of Kelvin.
-- [First-time setup](docs/setup.md) provides a guide on how to configure Kelvin for local development.
-- [Development](docs/development.md) contains useful hints on how to perform various
-  development tasks.
-- [Deployment](docs/deployment.md) describes how is Kelvin deployed to a production environment.
-
-You can find user documentation in the following files:
-- [Teacher's guide](docs/teacher-guide.md) contains useful information for teachers using Kelvin.
+You can find documentation of Kelvin's internal workings and how to contribute in [docs](https://mrlvsb.github.io/kelvin/).
```
Kobzol: Could you please add a command here that shows how to build the documentation locally?

Firestone82:
Interesting its not using .markdownlint.json file to disable some checks. But Im sure it was working fine. :|

.github/workflows/docs.yml
```diff
@@ -0,0 +1,81 @@
...
+      - ".github/workflows/docs.yml"
+
+  pull_request:
+    paths:
+      - "docs/**"
+      - ".github/workflows/docs.yml"
+
+  workflow_dispatch: {}
+
+permissions:
```
Kobzol: It's better to put permissions on jobs, to avoid jobs that run user code to not have elevated privileges. Could you please move this to the `deploy` job, please? Thanks.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,81 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Setup Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: 18
```
Kobzol: Why does this use a different Node version than in the `build` job? 🤔

Kobzol (commented):
Why are there duplicated .markdownlint files? I think that it would be enough if you first did `cd docs` and only then ran `markdownlint` on CI, to avoid this.

After you fix the nits, please also squash the commits. Thanks!

Kobzol:
Thank you! :)

PR #756, opened 2025-11-15
Deployment bugfixes, CI, concurrency

JersyJ:
- Increase timeout
- Move task to watch container logs after image pulling
- Fix deleting old image
- Disable concurrency
- Deployment Service CI
- Kelvin CI improvements

### DIFF
--- a/deployment_service/app/main.py
+++ b/deployment_service/app/main.py
@@ -15,19 +15,27 @@ app = FastAPI(
 deployment_lock = asyncio.Lock()
 
 
-async def check_deployment_concurrency():
-    """Dependency to ensure only one deployment runs at a time."""
+async def get_exclusive_deployment_access():
+    """
+    Dependency to ensure only one deployment runs at a time.
+    """
     if deployment_lock.locked():
         raise HTTPException(
             status_code=status.HTTP_429_TOO_MANY_REQUESTS,
             detail="A deployment is already in progress or was recently triggered. Please wait for it to complete before starting another one.",
         )
 
+    await deployment_lock.acquire()
+    try:
+        yield
+    finally:
+        deployment_lock.release()
+
 
 @app.post(
     "/",
     summary="Trigger Deployment for a specified service",
-    dependencies=[Security(validate_signature), Depends(check_deployment_concurrency)],
+    dependencies=[Security(validate_signature), Depends(get_exclusive_deployment_access)],
     response_model=DeploymentResponse,
     responses={
         status.HTTP_200_OK: {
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -1,78 +1,129 @@
...
+        uses: actions/checkout@v5
+
+      - name: Setup Python
+        id: setup_python
         uses: actions/setup-python@v5
         with:
-          python-version: "3.12"
+          python-version-file: "pyproject.toml"
+
+      - name: Install the latest version of uv
```
Kobzol: This name is misleading, because we install a pinned version, right? We shouldn't install the latest anyway, as `uv` isn't stable yet, and we don't want to upgrade `uv` and break our CI.

.github/workflows/ci.yml
```diff
@@ -1,78 +1,129 @@
...
+        uses: astral-sh/setup-uv@v7
+        with:
+          enable-cache: "auto"
+          version: "0.9.9"
+
+      - name: Install OS dependencies
+        run: |
+          sudo apt-get install -y libgraphviz-dev build-essential
+
+      - name: Cache venv
```
Kobzol: It takes ~25s to install dependencies. I'm pretty sure that caching them isn't worth it. Let's remove this.

.github/workflows/deployment_service-ci.yml
```diff
@@ -0,0 +1,219 @@
...
+        id: setup_python
+        uses: actions/setup-python@v5
+        with:
+          python-version-file: "deployment_service/pyproject.toml"
+
+      - name: Install the latest version of uv
+        uses: astral-sh/setup-uv@v7
+        with:
+          enable-cache: "auto"
+          working-directory: "deployment_service"
```
Kobzol: Same here, let's pin `uv` version.

.github/workflows/ci.yml
JersyJ: uv has advanced significantly over the past year and is now viewed as a stable, standard tool in the Python community. I’ll pin the version to keep things consistent for you.

.github/workflows/ci.yml
```diff
@@ -1,78 +1,129 @@
-name: CI
+name: Kelvin CI
 on:
   pull_request:
+      paths-ignore:
```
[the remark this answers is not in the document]
JersyJ: That doesn’t make sense. These are two separate services, each with its own CI. Each service might require a different setup and different steps—this isn’t programming, it’s setting up CI/CD. Also, in the context of Ruff, it could produce different results, with missing dependencies and settings.

.github/workflows/ci.yml
Kobzol: I meant stable as in semver. It frequently ships breaking changes - every 0.x (most recently 0.9) is potentially breaking and could break our CI.

JersyJ:
> I see the deployment service as part of the Kelvin codebase, and not something separate. I don't want to maintain two separate workflow files that do almost the same thing.

Ahh now I see that we’re looking at those components differently. From my perspective, your approach introduces constraints by tying everything to Kelvin. Sure, I can do that, but imagine the CI breaks e.g. because of the deployment_service while you’re focused on Kelvin and don’t want to touch the deployment part, it would still block you.

My approach treats it as a fully separate service that you only interact with when you actually want to modify it. It won’t get in your way if something breaks or becomes outdated, since it has no tight dependencies and can be unplugged whenever needed.

With that in mind, I’d like you to consider this direction. For me, the second option is much better, and maintaining it will cost roughly the same. I’ll also try to simplify it further.

Kobzol:
The deployment service is Kelvin-specific, and Kelvin currently can't be easily deployed without it, so Kelvin depends on it. I would prefer keeping everything together.

Kobzol:
I mean, we can definitely use it for more stuff within Kelvin itself, e.g. for evaluators, that's fine. But I don't want to maintain a service that would be usable outside of Kelvin. Of course, if you want to make it a separate thing, you can, that's fully up to you! But that ofc shouldn't live in this repository and should be its own separate thing.

.github/workflows/ci.yml
```diff
@@ -126,30 +231,54 @@ jobs:
...
+          fetch-depth: 0
           sparse-checkout: |
             deployment_service/deploy.py
           sparse-checkout-cone-mode: false
 
+      - name: Detect deployment-related changes
+        id: deployment-changes
+        shell: bash
+        run: |
+          BEFORE="HEAD^"
```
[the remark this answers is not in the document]
JersyJ: The Marketplace action is more mature, so I will change it.

I see you are a fan of GitHub-hosted runners when the bill is on GitHub, as this is a public repository.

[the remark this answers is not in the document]
Kobzol: Ok, I checked and we do disable squash merges and rebase merges, so in that case it should be fine, as long as the merge commit always has the previous merge commit as the *first* parent.

Could you please use `^1` to make that explicit, and add a comment that explains that we are comparing with the first parent of the current merge commit, which should be the previous upstream merge? Thanks!

Kobzol (approved):
Thank you! I'll merge it and deploy the new deployment service on the server, and then try to deploy some Kelvin version.

PR #769, opened 2025-11-22
Convert comments endpoint to API v2

Firestone82:
This PR changes endpoint `/task/<assignment_id>/<student_id>/<submit_id>/comments` to new v2 API and adds data classes. Also prepares API for next endpoints to come.

### DIFF
--- a/templates/web/submits_show_source.html
+++ b/templates/web/submits_show_source.html
@@ -23,7 +23,7 @@
         <div class="comments">
             <kelvin-submit-sources
                     url="{% url 'v2:retrieve_submit_details' submit.assignment.id submit.student submit.submit_num %}"
-                    commentUrl="{% url 'v2:create_submit_comment' submit.assignment.id submit.student submit.submit_num %}"
+                    comment_url="{% url 'v2:create_submit_comment' submit.assignment.id submit.student submit.submit_num %}"
             ></kelvin-submit-sources>
         </div>
       </div>
--- a/frontend/src/Comment.svelte
+++ b/frontend/src/Comment.svelte
@@ -11,7 +11,6 @@ export let text;
 export let type;
 export let id;
 export let can_edit;
-export let url = null;
 export let unread = null;
 export let notification_id = null;
 export let meta = null;


common/services/comments_service.py
```diff
@@ -0,0 +1,123 @@
+from typing import List
```
Kobzol: The "service" part is just noise. Let's just create `comments.py` and `submits.py` under the `common` directory.

api/v2/task/submit/comment.py
```diff
@@ -0,0 +1,68 @@
...
+):
+    comment: Comment = create_submit_comment(
+        submit=request.submit_instance,
+        author=request.user,
+        content=comment_data.text,
+        source=comment_data.source,
+        line=comment_data.line,
+    )
+
+    return to_dict(comment.to_dto(can_edit=True, type=comment.type(), unread=False))
```
Kobzol:
```suggestion
    return to_dict(comment.to_dto(can_edit=True, type=comment.type(), unread=True))
```
Otherwise this would be a behavior change.

api/auth.py
```diff
@@ -66,3 +70,35 @@ def wrapper(request, submit_id: int, *args, **kwargs):
         return function(request, submit_id, *args, **kwargs)
 
     return wrapper
+
+
+def require_submit_access(function):
```
Kobzol: Actually, this is too magical. Adding stuff onto `request` is very non-obvious. Let's just create a function that will return a submit and validate access, and use that in the API endpoints, something like this:
```python
submit = get_submit_write_access(assignment_id, login, submit_num)
```

api/v2/task/submit/comment.py
```diff
@@ -0,0 +1,68 @@
...
+
+@router.put("{comment_id}", url_name="modify_submit_comment")
+@require_submit_access
+def api_modify_submit_comment(
+    request,
+    assignment_id: int = Path(...),
+    login: str = Path(...),
+    submit_num: int = Path(...),
+    comment_id: int = Path(...),
+    comment_data: SubmitCommentUpdate = Body(...),
```
Kobzol: Same here, move the type above the function.

api/v2/task/submit/default.py
```diff
@@ -0,0 +1,79 @@
...
+)
+from web.dto import SubmitData
+from web.views.student import get_submit_data
+
+router = Router()
+
+
+@router.get("/details", url_name="retrieve_submit_details")
+@require_submit_access
+def api_get_submit_details(
```
Kobzol: There was no `/details` endpoint before, right? What is this?

Kobzol (commented):
Sorry for the delay, and thank you for improving this part of the codebase.

Reviewing such a large pull request, which combines functional and non-functional changes, moves stuff around, has a ~700 line diff, and is all within a single commit, is very difficult. It would be much better if this was split into multiple commits that would do self-contained atomic changes.

I left some comments, but it's not the whole feedback. If it could be split into smaller changes, that would make the review process feasible...

api/v2/task/submit/comment.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+@router.post("", url_name="create_submit_comment")
+@require_submit_access
+def api_create_submit_comment(
+    request,
+    assignment_id: int = Path(...),
+    login: str = Path(...),
+    submit_num: int = Path(...),
+    comment_data: SubmitCommentCreate = Body(...),
```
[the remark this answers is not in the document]
Kobzol: I would say that a "big file" is a file that has maybe more than 3 thousand lines of code, but even that doesn't necessarily has to be a problem. The issue is not size per-se, but mixing unrelated things together. If a file has thousands lines of code, but they are all related and have high cohesion, then it's not a problem. Sometimes there's just a lot of code :)

The DTO struct *is* related to the endpoint, and *only* to it, so there's no reason for it to be in a separate file. If it was used in multiple places, by different pieces of code, then maybe, but this is clearly a structure that is only used by a single endpoint, and thus it should be next to it, to make it clear what data does the endpoint receive without going to a separate file. I would like to move stuff out of `dto.py`, where it makes sense.

What we are discussing here is "organize code by feature vs by type" (e.g. https://softwareengineering.stackexchange.com/questions/338597/folder-by-type-or-folder-by-feature). Both can make sense, and Django clearly defaults to organizing by type, but I'm not a huge fan of that :) So if it's possible and it makes sense, I would rather organize stuff by feature.

api/v2/task/submit/comment.py
```diff
@@ -0,0 +1,68 @@
...
+@require_submit_access
+def api_modify_submit_comment(
+    request,
+    assignment_id: int = Path(...),
+    login: str = Path(...),
+    submit_num: int = Path(...),
+    comment_id: int = Path(...),
+    comment_data: SubmitCommentUpdate = Body(...),
+):
+    comment: Comment = update_submit_comment(
```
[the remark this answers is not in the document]
Kobzol: The function literally has `return None` in it, so the return type of the function is `Comment | None`, not `Comment`.

Kobzol:
It seems that modifying summary comments doesn't work.

Firestone82:
> image
> Looks like the POST method is unexpected by the endpoint?

I’m co fused, how did you get to that call? POST and PATCH was previously used for comments, but now it has separated endpoint

templates/web/submits_show_source.html
```diff
@@ -20,7 +20,12 @@
     </div>
     <div id="submit-{{ submit.assignment.id }}-{{submit.student}}" class="collapse">
       <div class="card-body">
-        <div class="comments"><kelvin-submit-sources url="{% url 'submit_comments' submit.assignment.id submit.student submit.submit_num %}"></kelvin-submit-sources></div>
+        <div class="comments">
+            <kelvin-submit-sources
+                    url="{% url 'v2:retrieve_submit_details' submit.assignment.id submit.student submit.submit_num %}"
+                    commentUrl="{% url 'v2:create_submit_comment' submit.assignment.id submit.student submit.submit_num %}"
```
Kobzol:
```suggestion
                    comment_url="{% url 'v2:create_submit_comment' submit.assignment.id submit.student submit.submit_num %}"
```
Otherwise the prop ends up being `undefined`.

PR #770, opened 2025-11-23
Fix LLM translation from english

Firestone82:
Currently it was thinking, that prompt is written in default language. As if it was czech, but prompt is in english.

### DIFF
--- a/common/ai_review/llm_reviewer.py
+++ b/common/ai_review/llm_reviewer.py
@@ -30,16 +30,13 @@ def remove_html_entities(text: str) -> str:
     return html_entity_pattern.sub("", text)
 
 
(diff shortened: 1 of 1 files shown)


common/ai_review/llm_reviewer.py
```diff
@@ -80,7 +77,8 @@ def process(self) -> AIReviewResult:
             ChatCompletionUserMessageParam(content=self.build_user_content(), role="user"),
         ]
 
-        if self.language.lower() != AI_REVIEW_DEFAULT_LANGUAGE:
+        # Since prompt is written in english, we need to translate when lnaguage not chosen as EN
```
Kobzol:
```suggestion
        # Since prompt is written in English, we need to translate when the output language is not English
```

PR #781, opened 2025-11-26
Fix missing unlink in cargo pipeline

Jan1s2:


evaluator/images/cargo/entry.py
```diff
@@ -195,11 +195,18 @@ def run_cargo(command: str, args: List[str]) -> BuildResult:
...
 
 try:
     result = run_cargo(COMMAND, ARGS)
 except BaseException as e:
     result = BuildResult.fail(f"Cargo execution failed\n{e}")
+finally:
+    for file in SANITIZED_FILES:
+        try:
+            os.unlink(file)
+        except FileNotFoundError:
```
Kobzol: Let's use `# noqa` here, same as in the `gcc` pipeline, and catch everything, to make sure that all files are unlinked.

Kobzol (approved):

PR #785, opened 2025-11-28
Basic seeding of a databaset for tasks and classes

Dym03:
This is a basic seeding of a database, that could lead to easier first time setup for new contributors, or for local testing of features. Right now this only seeds subject, classes, task and assigned tasks, also for the ease of use creates an teacher, to whom all of the created entities are assigned.
If this would be merged PR#525 could be closed as well, as it tries to solves the same issue.

Example of usage
```
python manage.py init_db
```

### DIFF
--- a/common/management/commands/init_db.py
+++ b/common/management/commands/init_db.py
@@ -1,0 +1,226 @@
+from dataclasses import dataclass
+from django.core.management.base import BaseCommand
+from django.contrib.auth.models import Group
+from common.models import Class, Subject, User, Semester, Task, AssignedTask
+from django.utils import timezone
+import datetime
+import os
+import random
+import string
+import logging
+from textwrap import dedent
+
+
+@dataclass
+class FileToCreate:
+    name: str
+    contents: str
+
+
+def create_task_files() -> list[FileToCreate]:
+    files = []
+    file_names = [".taskid", "config.yml", "readme.md"]
+    for file_name in file_names:
+        contents = ""
+        if file_name == "config.yml":
+            contents = dedent("""
+                # https://mrlvsb.github.io/kelvin/teachers-guide/task-configuration/
+                # You can also use CTRL+Space for autocompleting
+                pipeline:
+                # compile submitted source codes
+                - type: gcc
+                # flags: -Wall -Wextra -g -fsanitize=address -lm -Wno-unused-variable
+
+                # add hints from clang-tidy as comments
+                #- type: clang-tidy
+
+                # run tests
+                #- type: tests
+
+                # run custom commands
+                #- type: run
+                #  commands:
+                #    - ./main 123 | wc -l
+
+                # automatically assign points from the test results
+                #- type: auto_grader
+
+                #async:
+                # LLM is module for automatic feedback generation using Large Language Models
+                #llm:
+                #enabled: false
+                #language: cs
+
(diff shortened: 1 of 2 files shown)


Firestone82:
Looks nice and needed, for me it lacks few things, that are just my opinion as kelvin contributor.

1. It could also fill some data to config.yaml or README.md for those created tasks, so they are not empty. This can be just simple lorem for README.md, and config.yaml with some defaults.

2. Adding some comments could clearify those nested loops. That whole code is kinda too mutch "blocky" for me. I hate seeing function that have no spaces and its all mashed together.

3. Also some function for attaching those AssignedTask, would made that function bit cleaner since its already 80 lines long, with many inner loops.

Anyway, gj.

Dym03:
All of the above seems reasonable. Thanks for the feedback. But as a user, that haven't created any real tasks, do you maybe have a example of the config.yaml, or maybe do you have any good sources that would inspire me please.

Firestone82:
> All of the above seems reasonable. Thanks for the feedback. But as a user, that haven't created any real tasks, do you maybe have a example of the config.yaml, or maybe do you have any good sources that would inspire me please.

You can use default, that is generated once you click on its creation.
https://github.com/mrlvsb/kelvin/blob/8a949f738479c483c3d61a15217b29911a3c5f90/frontend/src/Manager.svelte#L51-L85

Firestone82:
Also how to configure, can be found at https://mrlvsb.github.io/kelvin/teachers-guide/task-configuration/

Dym03:
Tried to make the code clearer, added argument so that user can create one superuser and then also use the user as the teacher. Also updated the docs to add the step into the initial pipeline.

Firestone82 (approved):
Looks mutch better, atleast for me. But Im pretty sure, that @Kobzol will find something else :D Gj

common/management/commands/init_db.py
```diff
@@ -0,0 +1,239 @@
...
+
+@dataclass
+class FileToCreate:
+    name: str
+    contents: str
+
+
+def create_task_files() -> list[FileToCreate]:
+    files = []
+    file_names = [".taskid", "config.yaml", "readme.md"]
```
Kobzol:
```suggestion
    file_names = [".taskid", "config.yml", "readme.md"]
```
We hardcode this filename in a bunch of places.

common/management/commands/init_db.py
```diff
@@ -0,0 +1,239 @@
...
+) -> Class:
+    created_class, _ = Class.objects.get_or_create(
+        code=code, teacher=teacher, semester=semester, subject=subject, day=day, time=time
+    )
+    return created_class
+
+
+def create_random_class(teacher: User, semesters: list[Semester], subjects: list[Subject]) -> Class:
+    code = f"{random.choice(string.ascii_uppercase)}/{''.join(random.choices(string.digits, k=2))}"
+    chosen_semester = random.choice(semesters)
```
Kobzol: If the RNG is unlucky, the person could end up with no classes in the currently active semester :) FWIW, I don't think we even need to do multiple semesters. Bootstrapping a single semester is enough, and it should simplify the code considerably. We could make the semester year/summer be a parameter, if someone needed to overwrite it.

Kobzol (approved):
Awesome, thank you! :)

PR #789, opened 2025-12-04
Edit task page rewritten to Vue

SimonAdamek:
This is a rewrite of Edit task  page into Vue.

Old svelte version is still present and can be accessed by modifying URL 
http://127.0.0.1:8000/task/edit/8/ -> http://127.0.0.1:8000/#/task/edit/8/ 

Main changes:

- load page using standard Django urls.py and templates
- add useSvelteStoreInVue.ts to connect Vue components to the existing Svelte store
- rewrite the entire editor component (the previous version was present but malfunctioning)

The rest is essentially a 1:1 rewrite of the old components, with minor changes to make them work in Vue and make it little more clean

I tested editing and creating tasks, file management, tests management, and other related functionality. Everything appears to work the same way as before.


frontend/src/components/Editor.vue
```diff
@@ -1,79 +1,68 @@
...
+ * @prop {boolean} wrap       - wrap long lines, default false
+ * @prop {Record<string, () => void>} extraKeys
+ * default keys are:
+ *    - Ctrl-Space   autocomplete
+ *    - F11          full screen
+ *    - ESC          escape full screen
+ *    - Tab          soft tab
+ *    You can add custom ones in form { 'key': function }
+ */
+const props = defineProps({
```
Kobzol: We use the type-based `defineProps` approach everywhere else in Kelvin (`defineProps<{...}>(...)`). Was there something wrong with the original `defineProps` call here?

frontend/src/components/Editor.vue
```diff
@@ -88,138 +77,105 @@ function toMode(filename: string) {
...
-    if (result.spellCheck !== undefined) {
-      editor.setOption('spellcheck', result.spellCheck);
-    }
+function setUpOptions() {
+  if (codeMirrorEditor.value) {
+    codeMirrorEditor.value.setOption('mode', toMode(props.filename));
+    codeMirrorEditor.value.setOption('readOnly', props.disabled);
+    codeMirrorEditor.value.setOption(
+      'gutters',
+      props.filename == '/config.yml' ? ['CodeMirror-lint-markers'] : []
```
Kobzol: We spent quite a lot of effort on the old editor to make it generic and not hardcode any filepaths in it. Could you please clarify a bit more about what was broken about it? Can we make smaller fixes to it to make it work?

Kobzol (commented):
Hi, thank you for working on this, it will be great once we can move all of this to Vue! :) This is a pretty big change, and I'd like to try to land it in smaller pieces. Since you found a way to have both the Vue and Svelte versions running side-by-side, do you think that we could land the Vue rewrites in smaller chunks, while keeping Svelte as the default, and then do the switch and remove Svelte once everything has been migrated? It's quite hard to review a 2k line diff PR.

Kobzol (commented):
Left some review comments.

Kobzol (commented):
Left some review comments.

SimonAdamek:
Hopefully, this will work. I've been testing it  quite a while, and at least for me, it's stable. I changed the way the redirects work; it should work better with Vue Router, and it's the more official approach. This should also resolve the occasional issues with duplicates. I’ve also fixed the redirect issue that occurred after deleting a task using it.

Kobzol:
Thank you for the fixes. It seems that duplication still doesn't work correctly:
image
Even after reload of the page, it tries to load data from `.../asset/readme.md`, instead of `.../readme.md` for some reason. But maybe it's because of my weird local symlink setup.

In any case, it seems to work based on my local testing. I don't have time to review the whole changeset in detail, but I think that most of the issues have been resolved and we can fix the rest once we find it in production.

Thank you very much for implementing this, it's a big step towards finally removing Svelte (essentially just the class detail is left and then Svelte will be gone).

SimonAdamek:
It must be those symlinks. I've tried the duplication function several times and it never resulted in an error. 

https://github.com/user-attachments/assets/a02aeb3f-4591-4508-9c8a-3fe080c23599

PR #790, opened 2025-12-05
Add task deleter

Jan1s2:
Added automated deletion of containers. 
It uses the evaluator queue and does not expect any long-running containers to be running on the machine where the worker runs.

### DIFF
--- a/common/apps.py
+++ b/common/apps.py
@@ -1,22 +1,5 @@
 from django.apps import AppConfig
-import logging
 
 
 class CommonConfig(AppConfig):
     name = "common"
-
-    def ready(self):
-        # due to how django loads models, we can't import this at the top of the file
-        from common.cron_creator import create_crontask
-
-        try:
-            # We want to delete containers older than 30 minutes every hour
-            create_crontask(
-                "deleter",
-                "evaluator.docker_container_cleanup.delete_old_containers",
-                cron_string="0 * * * *",
-                args=[("int", "1800")],
-            )
-        # Database is not running, skip creating the cron task
-        except Exception as e:
-            logging.warning(f"Could not create task: {e}")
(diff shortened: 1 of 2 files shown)


task_creator/apps.py
```diff
@@ -0,0 +1,20 @@
...
+
+        try:
+            create_crontask(
+                "deleter",
+                "evaluator.old_deleter.delete_old_containers",
+                "0 * * * *",
+                args=[("int", "1800")],
+            )
+        # Database is not running, skip creating the cron task
+        except Exception:
```
Kobzol: Let's log the error at least.

task_creator/cron_creator.py
```diff
@@ -0,0 +1,24 @@
+from scheduler.models.scheduled_task import CronTask, BaseTask
+from scheduler.models.args import TaskArg
+from django.contrib.contenttypes.models import ContentType
+
+
+def create_crontask(name, callable, cron_string, args=[], enabled=True, queue="evaluator"):
```
Kobzol:
```suggestion
def create_crontask(name: str, callable, cron_string: str, args=[], enabled=True, queue="evaluator"):
```
Could you please also add a documentation string that explains what this does?

evaluator/old_deleter.py
```diff
@@ -0,0 +1,13 @@
+import docker
+from datetime import datetime, timezone
+
+
+def delete_old_containers(n, socket_path="unix://var/run/docker.sock"):
```
Kobzol: Please also add a comment here that explains what this function does.

task_creator/apps.py
```diff
@@ -0,0 +1,20 @@
...
+
+    def ready(self):
+        # due to how django loads models, we can't import this at the top of the file
+        from task_creator.cron_creator import create_crontask
+
+        try:
+            create_crontask(
+                "deleter",
+                "evaluator.old_deleter.delete_old_containers",
+                "0 * * * *",
```
Kobzol: Please add a comment here that explains that this runs once per hour.

PR #812, opened 2026-01-11
TaskDetail task page rewritten to Vue 

Firestone82:
This PR contains rewrite of TaskDetail from svelte into Vue. I tried to do smaller PRs with each component rewriten but Im not able to manage it as Im skilless in this type of frontend shenanigans.

Therefore this PR is again quite big, but It should be near copy from Svelte. So checking functionality should be easy as comparing two files, I tried it myself and corrected everything that was not working by AI conversion.

So, @Kobzol please check functionality as you would know all nits a tricks kelvin has. 😅 

In future PRs I will try to cleanup and enchance this TaskDetail page. Now in full Vue without using svelte. :)

### DIFF
--- a/frontend/src/Quiz/Lists/QuizList.vue
+++ b/frontend/src/Quiz/Lists/QuizList.vue
@@ -244,8 +244,8 @@ const filterChanged = () => {
   <DataTable ref="dataTable" class="table table-striped" :columns="columns" :options="options">
   </DataTable>
   <div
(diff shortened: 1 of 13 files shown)


Kobzol (approved):
Thank you, great job! I slightly tweaked the padding of the comment badges, and used Claude to port the `defineProps` calls to the style we already used in existing Vue components in Kelvin.

Issue #774, opened 2025-11-23
Non Intuitive sortering of files in TaskDetail Component.


Ten-James:
# Motivation

As a user, I find it impractical that the displayed files are sorted first by type (video, image, other) and only then by name. For example, in our ZPG class we currently have to submit both a video and a supplementary text. Unfortunately, the text then ends up somewhere in the middle of the page, causing related items that belonged together to become separated.

# Technical 

At this moment the sorting implementation is written here kelvin/web/views/student.py at line 887 
```py
"sources": sorted(result.values(), key=lambda f: (priorities[f["type"]], f["path"])),
```

# Possible Implementations
It would therefore be appropriate at least to add a file extension to video items so that they are consistent with the others, and then either:

1. Visually separate the sections—videos, images, text files—so it is immediately clear that the items are grouped in this way.

2. Keep everything sorted by name, or alternatively introduce collapsible folders and visually distinguish them from regular files.

> Personally, I would prefer the second option, but anything would be more intuitive than the current solution.

PR #813, opened 2026-01-17
Add view mode switch when viewing submits

Firestone82:
This PR adds a new button to the task view toolbar that toggles the view mode between List and Folder.

By default, the initial mode is determined by the number of files. Once the user manually changes the mode, their preference is saved to local storage and persists across sessions.

image

Fixes: https://github.com/mrlvsb/kelvin/issues/774

### DIFF
--- a/frontend/src/Student/TaskDetail.vue
+++ b/frontend/src/Student/TaskDetail.vue
@@ -1,5 +1,5 @@
 <script lang="ts" setup>
-import { computed, onMounted, onUnmounted, ref, watch } from 'vue';
+import { computed, onMounted, onUnmounted, ref } from 'vue';
 import SyncLoader from '../components/SyncLoader.vue';
 import SummaryComments from '../components/submit/SummaryComments.vue';
 import SubmitsDiff from '../components/submit/SubmitsDiff.vue';
@@ -597,28 +597,6 @@ onMounted(() => {
 onUnmounted(() => {
   window.removeEventListener('keydown', keydown);
   window.removeEventListener('hashchange', updateSelectedFileAndRows);
-});
-
-watch([files, viewModeValue], () => {
-  if (viewModeValue.value !== ViewModeState.TREE) {
-    return;
-  }
-
-  if (!files.value || files.value.length === 0) {
-    return;
-  }
-
-  const hasSelection =
-    selectedFilePath.value &&
-    files.value.some((file) => file.source.path === selectedFilePath.value);
-
-  if (!hasSelection) {
-    const firstFileIndex = findFirstFileIndex();
-
-    if (firstFileIndex !== -1) {
-      setSelectedFile(files.value[firstFileIndex].source.path);
-    }
-  }
 });
 </script>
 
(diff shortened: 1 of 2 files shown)


Kobzol:
Looks really awesome :heart: Thank you for working on this.

Some UX comments (haven't taken a look at code yet):
- It would be nice to have the ability to collapse/expand directories in the tree view. And by default, I'd mark all of them as collapsed. Helps for uploads with a lot of directories and files.
- In list view, it doesn't really make sense to collapse the file content after clicking on the filename.
- The collapse/expand all files button could be visible even in the tree view, but there it could collapse/expand all directories instead.
- I think that we can show the list/tree view switcher always, even if there is only a single file. Although with a single file, it should still default to list view.
- A tricky question is which file to open by default in the tree view :) Maybe, and we'd force the user to select? Or we could select heuristically a file called readme[.md] (case insensitive)? It seems like currently it opens the first file (alphabetically?) at the root of the project, which is a bit weird.

Firestone82:
> Sorry, I wanted to say "in **tree** view collapsing the single file doesn't make sense".
> 
> Sorting in backend is fine, but then the frontend can select the opened file.

Everything should be implemented as I envisioned. Please take a look and let me know if anything needs to be changed or if I missed any issues.

frontend/src/Student/TaskDetail.vue
```diff
@@ -408,25 +567,58 @@ const load = async () => {
...
+  window.addEventListener('hashchange', updateSelectedFileAndRows);
 });
 
 onUnmounted(() => {
   window.removeEventListener('keydown', keydown);
-  window.removeEventListener('hashchange', goToSelectedLines);
+  window.removeEventListener('hashchange', updateSelectedFileAndRows);
+});
+
+watch([files, viewModeValue], () => {
```
Kobzol: What is this supposed to do?

frontend/src/Student/TaskDetailSidebar.vue
```diff
@@ -0,0 +1,481 @@
...
+const handlePointerMove = (event: PointerEvent) => {
+  if (!isDragging) {
+    return;
+  }
+
+  const calculatedWidth = dragStartWidth + (event.clientX - dragStartX);
+  width.value = Math.min(Math.max(calculatedWidth, minWidth), maxWidth);
+};
+
+const stopDragging = () => {
```
Kobzol: What is this dragging thing?

Kobzol:
> > The empty file comments do not work, but I found out now that it is a pre-existing issue. Interesting that it works for suggestions though :)
> 
> Im not really sure if I understand. Whats the problem here?

I create a comment, it appears. Then I reload the page, and the comment is gone :) The comment counter shows it, but it is not rendered.

frontend/src/Student/TaskDetailSidebar.vue
```diff
@@ -0,0 +1,481 @@
...
+
+const isFolderCollapsed = (folderPath: string | null | undefined) => {
+  if (!folderPath) {
+    return false;
+  }
+
+  return collapsedByPath.value[folderPath] ?? false;
+};
+
+const allFoldersOpen = computed(() => {
```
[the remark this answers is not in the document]
Firestone82: True, will fix that.

frontend/src/Student/TaskDetailSidebar.vue
Kobzol: Ah, I noticed this, very cool! Was just confused about the name "dragging", I thought it's "drag and drop". Nevermind.

Issue #698, opened 2025-09-10
Add option to downoad individual files in submit


geordi:
It would be nice to have an option to download individual files in submit detail. Many times, there's just one file and downloading a submit creates a `tar.gz` file that has to be handled.

PR #815, opened 2026-01-19
Bring back download button for source

Firestone82:
This PR brings back previously removed download button. Also will resolve https://github.com/mrlvsb/kelvin/issues/698#issue-3402658860

Previously button was doing backend API call, but I dont really see why it was implemented that way, when on frontend we already have whole file as plain text.

Fixes: https://github.com/mrlvsb/kelvin/issues/698

### DIFF
--- a/common/submit.py
+++ b/common/submit.py
@@ -179,16 +179,14 @@ def fetch_submit_sources(submit: Submit) -> SubmitSources:
         # Text / fallback handling
         else:
             content_text: str = ""
-            content_url: str | None = None
             content_error: str | None = None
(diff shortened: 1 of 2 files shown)


Firestone82:
Ah I see, you right. So I will convert it to API call, thats not a big issue.

How should that "Large files" really work on tree view? For what can I say now, its using same logic as list view, so there will be just text "Content too large, show <a :href="file.source.content_url">raw content</a>."

Kobzol:
Yeah it should work like that. I just haven't tried it yet :)

PR #816, opened 2026-01-19
Add toast as global component and expose functionality

Firestone82:
This PR adds global Toast component that can be easily used in FE to show action response. 

https://github.com/user-attachments/assets/5364dd9e-e742-425a-9c31-0c0c9444de90

### DIFF
--- a/frontend/src/components/Toast.vue
+++ b/frontend/src/components/Toast.vue
@@ -1,89 +1,13 @@
-<script setup lang="ts">
-import { onBeforeUnmount, onMounted, ref } from 'vue';
-import type { ToastStatus, ToastOptions } from '../types/Toast';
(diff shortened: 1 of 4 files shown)


Firestone82:
> Having a global toast system is cool, but accessing it through `window` is like going back 15 years of frontend development 😅 Is there any specific reason for that? Should be enough to just import the toast functions as normal TS functions?

The only current reason is, that kelvin is still in Vue/Svelte state. So unless Im sure, that this toast methods won't be required in svelte. I can turn it into only TS import functions.

On the other hand, FE should not have any more svelte additions. So unless this functionality would be added into existing svelte. I can then accept not using window as cross-platform resolver.

Kobzol:
Yeah please make it Vue only :)

PR #817, opened 2026-01-19
Disable TREE view when opened on mobile.

Firestone82:
Quick to previous PR #813, that its sidebar is not greate to have on mobile devices.


Kobzol (approved):
Thank you for the quick fix! :)

PR #821, opened 2026-01-25
Refactor UploadSolution to Vue and fix unwanted image upload

Firestone82:
This PR migrates UploadSolution from Svelte to Vue and fixes an issue where images were unintentionally uploaded when dragged into the browser.

### DIFF
--- a/frontend/src/components/UploadSolution.vue
+++ b/frontend/src/components/UploadSolution.vue
@@ -237,9 +237,26 @@ const setupUploadInput = () => {
   fileInputLabel.value = input.nextElementSibling as HTMLElement | null;
 
   const handleChange = () => {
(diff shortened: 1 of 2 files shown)


Kobzol:
Thanks. While I was testing this change, I found something in the code which I didn't see before when you changed the submit page to Vue. The `remainingMS` mechanism, and some form of an upload queue (?) seems to be there, but it's broken. The `remainingMS` value is *always* negative, because it stores now() when a submit happens, and then it subtracts the current date from this value, so it can never be positive or even zero. There are a bunch of checks in the code that check this value and do something with an upload queue based on it. What is this about?

Given that it's broken and doesn't do anything, it would be great to remove it (in a follow-up PR)...

PR #824, opened 2026-02-09
Evaluator Images CI, Overall CI Improvements

JersyJ:
- CI for Evaluator Docker Images with build.py (Unable to use GitHub Action Cache due to docker-container driver design/architecture (https://github.com/moby/buildkit/issues/2343), reverting to local docker driver for compatibility.)
- Update CI to use native [GitHub Actions Cache](https://docs.docker.com/build/cache/backends/gha/) instead of the Registry cache, due to maintenance of the registry storage (LRU is used there automatically) and also the readability/visibility of that registry. Also this is official way and recommendation from GitHub and Docker
- Updated pyproject.toml dev dependency group (removed deprecated syntax)
- Migrate from pre-commit to Rust re-implementation called `prek`
- Update documentation and fix instructions that no longer works

### DIFF
--- a/evaluator/images/build.py
+++ b/evaluator/images/build.py
@@ -88,8 +88,28 @@ class ImageBuilder:
             else:
                 subprocess.check_call(cmd, cwd=image_dir)
 
-    def build(self, dry_run: bool = False):
+    def _get_all_parents(self, image: str) -> set[str]:
+        """Returns all transitive kelvin/ parents of the given image, including itself."""
+        result = {image}
+        queue = [image]
+        while queue:
+            current = queue.pop()
+            for parent in self.deps.get(current, []):
+                if parent.startswith("kelvin/") and parent not in result:
+                    result.add(parent)
+                    queue.append(parent)
+        return result
+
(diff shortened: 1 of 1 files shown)


evaluator/images/build.py
```diff
@@ -1,55 +1,109 @@
...
+    def build(self, dry_run: bool = False):
+        for batch in self.get_build_order():
+            for image in batch:
+                self._build_single_image(image, dry_run)
+
+
+def main():
+    parser = argparse.ArgumentParser(description="Manage Kelvin Docker images")
+
+    parser.add_argument("--dry-run", action="store_true", help="Print commands without executing")
```
Kobzol: Locally I often want to build only a specific image, to test something. Could you please add an optional argument that will build only one image with all its parents?

Kobzol (commented):
Thank you! Left one "feature request", but it doesn't have to be in this PR. Let me know if you wanna work on it or not, and if yes, if you wanna do it here or in another PR (in that case I'd merge this).

Kobzol (approved):
Great, thank you! Let's try this.

PR #830, opened 2026-02-17
Evaluator Rewrite - GCC

JersyJ:
- Rewrite of the GCC pipeline, to use entry.py in Evaluator and usage of `docker run` architecture instead of relying on `entry.py` in Docker Image
- Standardize the HTML output
- Use nh3 instead of bleach for sanitization
- Implements filename filtering to block files with dangerous characters from uploads and source listings


JersyJ:
Hi @Jan1s2, could you check this PR based on the previous PR discussion? Thanks a lot in advance.

PR #831, opened 2026-02-19
Evaluation refactoring (step 1): refactor config loading

Kobzol:
This PR refactors how the task evaluation context is loaded from disk. The previous code was hard to understand, and did a bunch of magic and stateful stuff. After the refactoring, it should hopefully be much clearer what is going on here.

I also changed the naming, which I plan to use more extensively in future PRs.

Pipeline will be named Workflow, and a single pipe will be called Job (to mirror terminology from GitHub Actions).

I also did some targeted compatibility breaks:
- It is no longer possible to load custom Pipe/Job implementations from the `script.py` file. I had no idea that something like this is even possible :D It was used by Dan in one of his tasks. I checked that the task should not be used anymore (and doesn't even work today anyway), and haven't found any other tasks on the server that would use this.
- It is no longer possible to specify `files` in the `config.yml -> tests` field nor in `tests.yml`. It was not used in any existing tasks, except for one of Dan's very old tasks.

This change also fixed a test ordering bug, which was not ordering tests correctly according to the order in `tests.yml`, and removes output of missing `Test` attribute in `task_detail.html` (the Test instance has no `assignment` field, no idea what that was).

I saw that many tasks use the `EvaluationContext` in `script.py::gen_tests`, and they access the `meta` field and the `create_test` method. Ideally, we should be only passing something with a very limited API to the scripts to avoid them accessing *whatever they want*, but for now I kept backwards compabitility to avoid too many changes at once.

### DIFF
--- a/evaluator/evaluation.py
+++ b/evaluator/evaluation.py
@@ -440,7 +440,7 @@ def record_test_files(tests: Dict[str, Test], task_path: str):
         name = f.split(".")[0]
         path = os.path.join(task_path, f)
 
-        n = f[len(name) + 1:]
+        n = f[len(name) + 1 :]
         if n in ["in", "out", "err"]:
             get_test(name).files["std" + n] = TestFile(File(path), n == "in")
 
--- a/common/evaluate.py
+++ b/common/evaluate.py
@@ -71,7 +71,7 @@ def evaluate_submit(request, submit, meta=None):
     meta = {
         **get_meta(submit.student.username),
         "before_announce": not is_teacher(submit.student)
-                           and submit.assignment.assigned > timezone.now(),
+        and submit.assignment.assigned > timezone.now(),
         "deadline": submit.assignment.deadline,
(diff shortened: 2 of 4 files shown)


evaluator/evaluation.py
```diff
@@ -182,146 +199,252 @@ def required_files(self):
...
+        return WorkflowConfigParseResult(
+            config=WorkflowConfig(queue=queue, timeout=timeout, tests=tests, jobs=jobs),
+            unknown_keys=unknown_keys,
+        )
+
+
+@dataclasses.dataclass()
+class WorkflowConfigParseResult:
+    config: WorkflowConfig
+    unknown_keys: list[str] = dataclasses.field(default_factory=list)
```
geordi: I think, I remember, we decided to keep using `List` and related types from `typing`. So I assume, we can use `list` from now on (or it may happed without me noticing :-) ).

PR #787, opened 2025-11-29
Rooms

geordi:
Let's start exam signup by adding a notion of rooms into the Kelvin.

PR #841, opened 2026-03-22
Ip filtering vue

SimonAdamek:
Allow submission uploads and task viewing only from defined classroom IP ranges

**Models**

- Added a new RoomIpRange model
- A Room can have multiple IP ranges

**Edit Task Page**

- Added a new field displaying a list of rooms (only for exam task type)
- The list is searchable and sorted by checked/unchecked status, then alphabetically

**IP-Based Access Control**

- Added a decorator to common/utils.py 
- used in web/views/student.py endpoints
- API also validates client IP addresses to prevent direct POST requests from unauthorized sources

**Administrator Page**

- Added a custom form to the Room model page
- Allows adding classrooms using IP ranges or CIDR notation

Basically same as #724 only based on new Vue.JS EditTask page.

First commit is 457d3fa9b8860d6e4e434e05852fa3f15efca4a0

img3
img2
img1

Related to #586


Firestone82:
I can smell that "Too big PR" from @Kobzol around the corner. 👀
But I had just a small look, and it looks like you kinda mixing naming and utility usage in frontend.

You have written new `utilities/useSvelteStoreInVue.ts` when tehre is already `utilities/useSvelteStore.ts`. So this should be merged together?

And also folder naming should be somehow identical, so there is no kebab and camel case next to each other.

SimonAdamek:
@Firestone82  

That "Too big PR" is actually this one: https://github.com/mrlvsb/kelvin/pull/789. This one is built on top of it, which is why, before merge, both commits were displayed here, making it look even bigger...

common/models.py
```diff
@@ -236,6 +236,15 @@ class Meta:
         verbose_name_plural = "classes"
 
 
+class ClassroomIpRange(models.Model):
```
Kobzol: I think you missed https://github.com/mrlvsb/kelvin/pull/787? We already have models for rooms and for `RoomIpRange`. Could you please base your work on top of that? Should hopefully be a small-ish change.

common/submit.py
```diff
@@ -96,6 +101,12 @@ def store_submit(request: HttpRequest, assignment: AssignedTask) -> Submit:
     if client_ip:
         s.ip_address = client_ip
 
+    if assignment.allowed_classrooms:
```
Kobzol: The logic about empty `allowed_classrooms` should be within `is_allowed_from_ip`, so that callers of this method cannot forget about performing this check.

common/migrations/0030_classroomiprange_assignedtask_allowed_classrooms.py
```diff
@@ -0,0 +1,27 @@
+# Generated by Django 4.2.16 on 2025-10-23 12:06
```
Kobzol: The migrations should be reworked; delete all of the added migrations and generate them afresh from the latest version of the code.

common/models.py
```diff
@@ -137,6 +138,17 @@ def __str__(self):
         return f"{self.code} - {self.capacity} seats"
 
 
+class RoomIpRange(models.Model):
+    ip_range_start = models.GenericIPAddressField()
+    ip_range_end = models.GenericIPAddressField()
+    room = models.ForeignKey(
+        Room, on_delete=models.SET_NULL, blank=True, null=True, related_name="ip_ranges"
```
Kobzol:
```suggestion
        Room, on_delete=models.CASCADE, null=False, related_name="ip_ranges"
```

common/models.py
```diff
@@ -257,6 +270,23 @@ def is_past_deadline(self):
             and datetime.datetime.now(datetime.timezone.utc) > self.deadline
         )
 
+    def is_allowed_from_ip(self, ip: str) -> bool:
+        ip = ipaddress.ip_address(ip)
```
Kobzol: This will raise `ValueError` if the IP address is invalid, and more importantly if it is `None`. I guess then we have to decide what do we do. A safer option is to reject, but that can cause some false positives, hmm.

Let's allow it for now (`return True`), but also log a warning that the IP address was not parsed correctly. Include the IP address content and the assignment name in the log message.

common/models.py
```diff
@@ -257,6 +270,23 @@ def is_past_deadline(self):
...
+
+        if not self.allowed_rooms.all().exists():
```
Kobzol: Let's eagerly fetch the rooms into a local list variable, otherwise every `.all()` call can create a separate SQL query. For bonus points, use https://docs.djangoproject.com/en/dev/ref/models/querysets/#prefetch-related to prefetch `ip_ranges`.

common/models.py
```diff
@@ -257,6 +270,23 @@ def is_past_deadline(self):
...
+            return True
+
+        allowed = False
+
+        for room in self.allowed_rooms.all():
+            for room_range in room.ip_ranges.all():
+                start = ipaddress.ip_address(room_range.ip_range_start)
+                end = ipaddress.ip_address(room_range.ip_range_end)
+
+                allowed |= start <= ip <= end
```
Kobzol: There is no need for `|=`. Once we find an allowed range, we can eagerly exit with `return True`. After the nested cycles, the function should have `return False`, and the `allowed` variable is not needed.

common/utils.py
```diff
@@ -92,6 +97,59 @@ def get_client_ip_address(request: HttpRequest) -> IPAddressString | None:
...
+        try:
+            assignment = AssignedTask.objects.get(pk=assignment_id)
+        except AssignedTask.DoesNotExist:
+            raise Http404(f"AssignedTask with id {assignment_id} not found")
+
+        # allow after deadline
+        if assignment.deadline is not None and timezone.now() > assignment.deadline:
+            return function(*args, **kwargs)
+
+        if assignment.allowed_rooms:
```
Kobzol, line 138: There is a missing `else` condition that forwards to `function(*args, **kwargs*)`, right?

common/utils.py
```diff
@@ -92,6 +97,59 @@ def get_client_ip_address(request: HttpRequest) -> IPAddressString | None:
...
+            x_forwarded_for = request.META.get("HTTP_X_FORWARDED_FOR")
```
Kobzol: Please use `get_client_ip_address` to find the IP address, so that we have unified logic for this everywhere.

api/views/default.py
```diff
@@ -732,6 +745,15 @@ def set_subject(task):
...
                     },
                 )
+
+                if "allowed_rooms" in cl:
+                    allowed_rooms = []
+                    for class_id in cl["allowed_rooms"]:
+                        room_object = Room.objects.get(pk=class_id)
+                        allowed_rooms.append(room_object)
+
+                    assigned_task.allowed_rooms.set(allowed_rooms)
```
Kobzol, line 755: Does this automatically update the object? Don't we have to call `assigned_task.save()` after?

common/utils.py
```diff
@@ -92,6 +97,59 @@ def get_client_ip_address(request: HttpRequest) -> IPAddressString | None:
         return IPAddressString(client_ip)
 
 
+def ip_address_check(function):
```
Kobzol:
```suggestion
def assignment_ip_check(function):
```

web/views/student.py
```diff
@@ -824,11 +826,13 @@ def submit_download(
 
 
 @login_required
+@ip_address_check
```
Kobzol: This shouldn't be here.

web/views/student.py
Kobzol, line 317: `submit_download` and `unmark_solution_final_mark` miss the decorator.

common/admin.py
```diff
@@ -140,6 +145,60 @@ class SubmitAdmin(admin.ModelAdmin):
     autocomplete_fields = ["assignment"]
 
 
+class RoomIpRangeAdminForm(forms.ModelForm):
+    use_cidr = forms.BooleanField(required=False, label="Enter CIDR instead of range")
+
+    ip_range_start = forms.GenericIPAddressField(required=False)
+    ip_range_end = forms.GenericIPAddressField(required=False)
+    cidr = forms.CharField(required=False, label="CIDR address (10.0.0.1/24")
```
Kobzol:
```suggestion
    cidr = forms.CharField(required=False, label="CIDR address (10.0.0.1/24)")
```

frontend/src/Teacher/EditTask/RoomsSelect.vue
```diff
@@ -0,0 +1,128 @@
...
+          :id="'classroom' + room.id"
+          :checked="room.isSelected"
+          type="checkbox"
+          class="form-check-input"
+          @change="() => toggleItem(room)"
+        />
+        <label class="form-check-label" :for="'classroom' + room.id">{{ room.code }}</label>
+      </div>
+
+      <small v-if="selectedCount === 0" class="text-black">No classrooms found</small>
```
Kobzol: This always shows "no classrooms found" when *no classroom is selected*, that doesn't make much sense 😅

frontend/src/Teacher/EditTask/RoomsSelect.vue
```diff
@@ -0,0 +1,128 @@
...
+
+const selectedRooms = defineModel<number[]>({
+  required: false
+});
+
+let allRooms = ref<Room[]>(null);
+
+const vClickOutside = clickOutside;
+
+onMounted(async () => {
```
Kobzol: If there will be 20 classes, then every one of them will load the classrooms list separately (and it also produces an annoying flash after the page is loaded :) ). Could you please load the list of classrooms once in `EditTask`, and then pass it as a prop to all `RoomsSelect` components?

frontend/src/Teacher/EditTask/RoomsSelect.vue
```diff
@@ -0,0 +1,128 @@
...
+interface ViewRoom extends Room {
+  isSelected: boolean;
+}
+
+let { onDuplicateClick, disabled = false } = defineProps<{
+  onDuplicateClick: (selected: number[]) => void;
+  disabled: boolean;
+}>();
+
+const selectedRooms = defineModel<number[]>({
```
Kobzol: Why `defineModel`? 🤔 We don't have two-way binding of this, right?

Kobzol (commented):
Thank you, this is really cool! I'm still not 100% sure if the decorator approach is the right one, and I'm sure we missed some endpoints, but let's try it. Left a bunch of comments on the implementation.

common/models.py
SimonAdamek: I assumed that IP addresses would only be entered into the system via the admin interface, where the system checks if the input contains a valid IP address. There is also a check for empty fields.

api/views/default.py
SimonAdamek: It works, and according to the documentation, that should be enough.

```
Note

Note that add(), aadd(), create(), acreate(), remove(), aremove(), clear(), aclear(), set(), and aset() all apply
database changes immediately for all types of related fields. In other words, there is no need to call save()
asave() on either end of the relationship....
```

[Many-to-many relationships](https://docs.djangoproject.com/en/dev/ref/models/relations/#django.db.models.fields.related.RelatedManager.set)

common/models.py
Kobzol: But in this method we check the IP that comes from a user HTTP request, not an IP configured in the admin interface.

api/views/default.py
Kobzol: Thanks!

web/views/student.py
SimonAdamek: I completely agree with `submit_download`, but `unmark_solution_final_mark` is only for teachers - no one else can remove the mark, so why should it be filtered?

Kobzol: Right, I forgot that only teachers can do that, nevermind then.

common/models.py
SimonAdamek: Oh, I'm sorry, you're absolutely right.

frontend/src/Teacher/EditTask/RoomsSelect.vue
SimonAdamek: The `selectedRooms` are stored in the `AssignedTask` object that is loaded and saved in the EditTask component. This is a two-way binding because we want to display the previously saved classes and save the new ones after they are selected.

Kobzol: Hmm, I don't like that :laughing: I'd rather store the source of truth at a single place (the parent), and then make the child component (`RoomsSelect`) just read the prop from the parent and update it via an event. Having a model here, especially one bound to the field of an object that comes from a remote API, feels very magical.

But I at least understand it now, thank you. Let's leave it as it is for now.

common/models.py
```diff
@@ -270,22 +268,27 @@ def is_past_deadline(self):
             and datetime.datetime.now(datetime.timezone.utc) > self.deadline
         )
 
-    def is_allowed_from_ip(self, ip: str) -> bool:
+    def assignment_ip_check(self, ip: str) -> bool:
+        print(self.allowed_rooms.exists())
```
Kobzol: Leftover print.

common/models.py
```diff
@@ -270,22 +268,27 @@ def is_past_deadline(self):
             and datetime.datetime.now(datetime.timezone.utc) > self.deadline
         )
 
-    def is_allowed_from_ip(self, ip: str) -> bool:
+    def assignment_ip_check(self, ip: str) -> bool:
```
Kobzol: `is_allowed_from_ip` was a good name, I wanted to rename `ip_address_check` to `assignment_ip_check`, the decorator 😅

common/models.py
Kobzol: This would still produce unnecessary queries:
Image

I tried with this:
```python
allowed_rooms_list = self.allowed_rooms.prefetch_related("ip_ranges").all()
        if not allowed_rooms_list:
            return True

        try:
            ip = ipaddress.ip_address(ip)
        except ValueError:
            logging.warning(
                "Invalid IP address: {} while attempting to access AssignedTask {}".format(
                    ip, self.task.name
                )
            )
            return True

        for room in allowed_rooms_list:
```

which seems to do the minimal amount of work required:
Image

Kobzol (commented):
Thank you for the edits, left a few more comments.

SimonAdamek:
GitHub tests are failing during the migration check, but I see in the log  

`performed for database connection 'default': connection failed: connection to server at "127.0.0.1", port 5432 failed: Connection refuse.
 Is the server running on that host and accepting TCP/IP connections?`

 This isn't my fault, is it?

Kobzol:
Oh, turns out that the CI error was actually caused by this PR. The migrations were not regenerated (the error message from `makemigrations --check` is quite confusing...).

Kobzol (approved):
Thank you! And sorry for the long review.

PR #842, opened 2026-03-22
Make .well-known directory accessible

DanielKrasny:
Even though `.well-known/security.txt` file was added (#760), it was never accessible on that path, not even locally. I added a static URL pattern to cover `.well-known` directory. For nginx, I added a new volume with static data and modified the config, since `collectstatic` won't include this file, no matter it's present in the static directory.

Additionally, I modified the content of `security.txt` file to comply with [RFC 9116](https://www.rfc-editor.org/rfc/rfc9116).

### DIFF
--- a/deploy/nginx.conf
+++ b/deploy/nginx.conf
@@ -48,6 +48,11 @@ http {
             gzip_vary on;
         }
 
+        location /.well-known {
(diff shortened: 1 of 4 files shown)


Kobzol (approved):
Thank you!

PR #845, opened 2026-03-31
Fix submit comments not updating value

DanielKrasny:
Some students reported that comments submitted via the Vue comment component are not sent because the API request sends an empty message. I noticed that `CommentForm` and `QuizEdit` components are the only ones using `v-model:value` for `Editor` component instead of `v-model`. Since we don't use multiple bindings here, I think the default one is sufficient and fixes the problem.


Kobzol (approved):
Thank you! :)

PR #850, opened 2026-04-12
Fix copy to clipboard tooltip showing correctly.

Firestone82:
Fixes copy to clipboard showing on new line instead next to it with fade effect.
Also removes duplicate class which was +- same.

https://github.com/user-attachments/assets/6f6a44ff-d4be-4a13-90f3-237554d693be


frontend/src/Teacher/EditTask/AutoCompleteTaskPath.vue
```diff
@@ -9,7 +9,7 @@ import { clickOutside } from '../../utilities/clickOutside';
 import { User } from '../../utilities/SvelteStoreTypes';
 import { useReadableSvelteStore } from '../../utilities/useSvelteStoreInVue';
 import { defineModel, onMounted, computed, ref, watch, defineEmits } from 'vue';
-import CopyToClipboard from './CopyToClipboard.vue';
```
Kobzol, line 12: I think that this file can be deleted now?

Kobzol (commented):
Thank you, this was bothering me 😆

frontend/src/Teacher/EditTask/AutoCompleteTaskPath.vue
Firestone82: It can't be deleted, because ist beeing used in html. That import was just moved.

Firestone82: Ah I see now, that previous class deletion is not in commit. 😅

Kobzol (approved):
Thank you! :)

PR #852, opened 2026-04-25
Fix ESLint warnings

DanielKrasny:
Another detail that annoyed me during development: there are a few files that produce warnings in ESLint, and they were even easy to fix just by running `npm run fix`. When I was into it, we have many places where we use `v-html` which is obviously considered insecure unless we know what will be in such elements or how we are using it. I suppressed those warnings, because they are reported in every single PR even though we know the use of `v-html` is intentional.

As a bonus, I noticed a small detail in the `Notifications` component, when there's no `action_object_url`, an `action_object` string should be displayed, but someone forgot to add double curly braces to make it work. Tests for `Notifications` component were also failing because of the ESLint warning suppression, so I made a new snapshot.

The entire frontend codebase is now properly formatted with Prettier and without ESLint warnings.

### DIFF
--- a/frontend/src/Quiz/Lists/QuizList.vue
+++ b/frontend/src/Quiz/Lists/QuizList.vue
@@ -265,8 +265,12 @@ const filterChanged = () => {
           <div class="col-12 mb-1">
             <label for="subject-select" class="form-label">Subject</label>
             <select id="subject-select" v-model="selectedSubjectAbbr" class="form-control">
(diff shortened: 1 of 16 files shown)


Kobzol (approved):
Thanks! camelCase is nicer.

Kobzol:
This broke creating and editing of comments. Seems like converting case is not very safe :)

PR #856, opened 2026-05-09
Update Python dependencies for minor versions & Formatting

JersyJ:
Preparation for enabling renovatebot


JersyJ:
Not needed anymore.

PR #858, opened 2026-05-10
Fix inability to edit comments

DanielKrasny:
The ESLint fix (#852) changed prop bindings from API types in comments by forcing them to `camelCase`, which caused issues with both creating and editing comments. After trying multiple approaches to keep the properties in `camelCase`, I decided to revert them back to `snake_case` and suppress the ESLint warning for now.

I also reviewed my previous PR once more to identify any other possible issues. The only additional problem I found was incorrect passing of the `quizId` prop in the quiz submit list template in Django.

*BTW: I’m really impressed that Rust can automatically convert `camelCase` to `snake_case` for de-/serialization simply by using `#[serde(rename_all = "snake_case")]`. Sadly, I couldn’t find an equally straightforward solution in TypeScript. It seems the only option is to transform object keys (deeply) at runtime. The question is: should I investigate this further so we can use `camelCase` properties throughout the frontend while keeping the API in `snake_case`?*

### DIFF
--- a/frontend/src/components/submit/CodeRow.vue
+++ b/frontend/src/components/submit/CodeRow.vue
@@ -75,13 +75,13 @@ const addNewComment = (text: string) => {
       <template v-for="comment in comments || []" :key="comment.id">
         <SuggestedComment
           v-if="comment.type === 'ai-review'"
-          v-bind="comment"
+          :comment="comment"
           @resolve-suggestion="emit('resolveSuggestion', $event)"
         />
 
         <Comment
           v-else
-          v-bind="comment"
+          :comment="comment"
           @save-comment="emit('saveComment', $event)"
           @set-notification="emit('setNotification', $event)"
         />
--- a/frontend/src/components/submit/Comment.vue
+++ b/frontend/src/components/submit/Comment.vue
@@ -6,30 +6,32 @@ import { safeMarkdown } from '../../markdown';
 import { hideComments, HideCommentsState } from '../../stores';
 import { useSvelteStore } from '../../utilities/useSvelteStore';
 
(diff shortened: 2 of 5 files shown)


Kobzol:
I don't think there is an easy way of automatically converting the case.

I see where the error originates now - the usage of `v-bind` is not what caused it. We should not be tying the casing on the frontend and on the backend, just because they use a subset of the same fields. That is what leads to similar problems. At the very least, we should not repeat the comment attributes in props of the comment component.

I would suggest either:
1) Remove usage of `v-bind="comment"` (there are 4 usages of it, I think), and take a `Comment` data type as a single input prop instead. The datatype will be defined as `type Comment = { ... }`, and its attributes will correspond to what is returned from the backend.
2) Remove usage of `v-bind`, and manually pass all required props to the comment components, like this: `<Comment notification-id="comment.notification_id" ...`, and then in `Comment` there would be `defineProps<{ notificationId...`.

I'd prefer variant 1.

DanielKrasny:
Okay, I agree that we shouldn't "improve" the casing of API responses, since after a few tests against the current codebase on my side, I found my test implementation of case conversion problematic, especially when dealing with built-in JS objects. So it is definitely better to keep the `snake_case` from the API and `camelCase` for the rest of the frontend whenever possible.

Regarding this PR, I have refactored the `Comment` component to accept the `Comment` API type as a single prop instead of duplicating the props individually, as you proposed in variant 1.

Kobzol:
Oh, there already was a type definition for a comment. Well, even better.

Thank you! Let's try.

Kobzol (approved):

PR #859, opened 2026-05-13
Show late submits in students view

Dym03:
If the first submit was after a deadline, user could see only a dash icon in the submitted column leaving him unsure if he submitted it or not. This PR add a simple red check icon, notifing the user he succesfully submited but after the set deadline. Maybe the different color is not needed, but wasn't sure about that.

Thought:
Go to latest submit in the first branch only takes you to the latest submit before deadline. Sure this is wanted but maybe it might be misleading. Maybe merging the 2 branches and just distinguish the late submits by the color, but jump really to the latest submit.

Current State:
Screenshot From 2026-05-13 19-19-57

### DIFF
--- a/templates/web/index.html
+++ b/templates/web/index.html
@@ -86,6 +86,10 @@
                                                title="Go to latest submit">
                                                <span class="iconify" data-icon="bx:bx-check"></span>
                                             </a>
+                                        {% elif task.submits %}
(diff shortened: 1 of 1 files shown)


Kobzol (approved):
Sorry for the delay in review, I didn't have much extra time in the recent months. Jumping to the latest submit before the deadline is more for the teacher, rather than the student, to ensure that you don't rate a late submitted commit by accident.

Thank you, this make sense!

PR #860, opened 2026-05-26
Fix light theme highlighting

Firestone82:
This PR fixes task config syntax hightlighting in white theme.

image
image


Kobzol (approved):
Seems messy, but I guess we can't do much better before Svelte is removed. Thank you!

PR #861, opened 2026-06-19
Run Postgres in CI job

Kobzol:
I have no idea how our migration check in CI worked without this before..


Kobzol:
Ah, I see. A live DB is not actually needed to check migrations.

